PR #1, opened 2017-05-25
Implement Atomic/Owned/Ptr

ghost:
This implementation is based on the original [RFC](https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-05-02-atomic-api.md).

A note on style:

The name of the crate is rather long (`crossbeam-epoch`), so calling methods from it is rather awkward. Consider:
```rust
extern crate crossbeam_epoch;

fn main() {
    crossbeam_epoch::pin(|scope| {});
}
```

I like aliasing the crate name like this:
```rust
extern crate crossbeam_epoch as epoch;
```

Or even when importing:
```rust
use crossbeam_epoch as epoch;
```

Examples in the documentation use the latter style.

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -266,11 +266,10 @@ impl<T> Atomic<T> {
         ord: Ordering,
         _: &'scope Scope,
     ) -> Result<(), Ptr<'scope, T>> {
-        let previous = self.data.compare_and_swap(current.data, new.data, ord);
-        if previous == current.data {
-            Ok(())
-        } else {
-            Err(Ptr::from_data(previous))
+        let fail_ord = strongest_failure_ordering(ord);
+        match self.data.compare_exchange(current.data, new.data, ord, fail_ord) {
+            Ok(_) => Ok(()),
+            Err(previous) => Err(Ptr::from_data(previous)),
         }
     }
 
@@ -350,13 +349,14 @@ impl<T> Atomic<T> {
         ord: Ordering,
         _: &'scope Scope,
     ) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)> {
-        let previous = self.data.compare_and_swap(current.data, new.data, ord);
-        if previous == current.data {
-            let data = new.data;
-            mem::forget(new);
-            Ok(Ptr::from_data(data))
-        } else {
-            Err((Ptr::from_data(previous), new))
+        let fail_ord = strongest_failure_ordering(ord);
+        match self.data.compare_exchange(current.data, new.data, ord, fail_ord) {
+            Ok(_) => {
+                let data = new.data;
+                mem::forget(new);
+                Ok(Ptr::from_data(data))
+            }
+            Err(previous) => Err((Ptr::from_data(previous), new)),
         }
     }
 
@@ -592,7 +592,7 @@ impl<T> Owned<T> {
     /// let o = unsafe { Owned::from_raw(Box::into_raw(Box::new(1234))) };
     /// ```
     pub fn from_box(b: Box<T>) -> Self {
-        unsafe { Self::from_raw(Box::into_raw(b)) }
+        Self::from_raw(Box::into_raw(b))
     }
 
     /// Returns a new owned pointer pointing to `raw`.
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -28,6 +28,13 @@ fn ensure_aligned<T>(raw: *const T) {
     assert!(raw as usize & low_bits::<T>() == 0, "unaligned pointer");
 }
 
+/// Panics if the tag doesn't fit into the unused bits of an aligned pointer to `T`.
+#[inline]
+fn validate_tag<T>(tag: usize) {
+    let mask = low_bits::<T>();
+    assert!(tag <= mask, "tag too large to fit into the unused bits: {} > {}", tag, mask);
+}
+
 /// Returns a bitmask containing the unused least significant bits of an aligned pointer to `T`.
 #[inline]
 fn low_bits<T>() -> usize {
@@ -38,9 +45,8 @@ fn low_bits<T>() -> usize {
 /// Panics if the tag doesn't fit into the unused bits of the pointer.
 #[inline]
 fn data_with_tag<T>(data: usize, tag: usize) -> usize {
-    let mask = low_bits::<T>();
-    assert!(tag <= mask, "tag too large to fit into the unused bits: {} > {}", tag, mask);
-    (data & !mask) | tag
+    validate_tag::<T>(tag);
+    (data & !low_bits::<T>()) | tag
 }
 
 /// An atomic pointer that can be safely shared between threads.
@@ -61,7 +67,7 @@ unsafe impl<T: Send + Sync> Send for Atomic<T> {}
 unsafe impl<T: Send + Sync> Sync for Atomic<T> {}
 
 impl<T> Atomic<T> {
-    /// Returns a new atomic pointer initialized with the tagged pointer `data`.
+    /// Returns a new atomic pointer pointing to the tagged pointer `data`.
     fn from_data(data: usize) -> Self {
         Atomic {
             data: AtomicUsize::new(data),
@@ -78,7 +84,25 @@ impl<T> Atomic<T> {
     ///
     /// let a = Atomic::<i32>::null();
     /// ```
+    #[cfg(not(feature = "nightly"))]
     pub fn null() -> Self {
+        Atomic {
+            data: AtomicUsize::new(0),
+            _marker: PhantomData,
+        }
+    }
+
+    /// Returns a new null atomic pointer.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_epoch::Atomic;
+    ///
+    /// let a = Atomic::<i32>::null();
+    /// ```
+    #[cfg(feature = "nightly")]
+    pub const fn null() -> Self {
         Atomic {
             data: AtomicUsize::new(0),
             _marker: PhantomData,
@@ -242,10 +266,11 @@ impl<T> Atomic<T> {
         ord: Ordering,
         _: &'scope Scope,
     ) -> Result<(), Ptr<'scope, T>> {
-        let fail_ord = strongest_failure_ordering(ord);
-        match self.data.compare_exchange(current.data, new.data, ord, fail_ord) {
-            Ok(_) => Ok(()),
-            Err(previous) => Err(Ptr::from_data(previous)),
+        let previous = self.data.compare_and_swap(current.data, new.data, ord);
+        if previous == current.data {
+            Ok(())
+        } else {
+            Err(Ptr::from_data(previous))
         }
     }
 
@@ -325,14 +350,13 @@ impl<T> Atomic<T> {
         ord: Ordering,
         _: &'scope Scope,
     ) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)> {
-        let fail_ord = strongest_failure_ordering(ord);
-        match self.data.compare_exchange(current.data, new.data, ord, fail_ord) {
-            Ok(_) => {
-                let data = new.data;
-                mem::forget(new);
-                Ok(Ptr::from_data(data))
-            }
-            Err(previous) => Err((Ptr::from_data(previous), new)),
+        let previous = self.data.compare_and_swap(current.data, new.data, ord);
+        if previous == current.data {
+            let data = new.data;
+            mem::forget(new);
+            Ok(Ptr::from_data(data))
+        } else {
+            Err((Ptr::from_data(previous), new))
         }
     }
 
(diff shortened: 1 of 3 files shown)


Vtec234:
LGTM!

Some things that might be worth adding:
- `const fn` versions of the `Atomic::null` and `Ptr::null` zero-initializers. The usage for this is that static contexts require `const` initialization. And when the [progress of this feature](https://github.com/rust-lang/rust/issues/24111) allows for constant arguments, we could extend this to `::new` and `::from_raw`.
- From [this PR](https://github.com/crossbeam-rs/crossbeam/pull/70/files), bitwise `fetch_or`, `fetch_and`, `fetch_xor` operations that act on the tag. Being able to do this instead of using a CAS turns out to be useful both for aesthetic reasons and practical ones, since `fetch_X` can potentially be more optimizable. An example usage is in lists, where we mark a pointer with `1` to indicate removal. The signatures would look like:
`fn fetch_or<'scope>(&self, tag: usize, ord: Ordering, _: &'scope Scope) -> Ptr<'scope, T> `
for all operations.
- More tests. There are currently some doctests for `Ptr`. We don't need to test the atomic operation proxy methods, since they delegate to stdlib and we trust that, but it would be good to test the additions - notably the pointer masks, e.g. `assert_eq!(low_bits::<u8>(), 0)`, `u16, 1` (is this guaranteed?), etc.

What does everybody think?

src/atomic.rs
```diff
@@ -0,0 +1,860 @@
...
+use std::ops::{Deref, DerefMut};
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+
+use Scope;
+
+/// Given ordering for the success case in a compare-exchange operation, returns the strongest
+/// appropriate ordering for the failure case.
+#[inline]
+fn strongest_failure_ordering(ord: Ordering) -> Ordering {
```
jeehoonkang, line 13: I would rather prefer introducing `compare_exchange` that exposes the ordering for failure cases.  For example, it would be useful to have a `compare_exchange` with acquire-release ordering when successful, and relaxed ordering when unsuccessful.

src/atomic.rs
```diff
@@ -0,0 +1,860 @@
...
+    /// Returns a new null atomic pointer.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_epoch::Atomic;
+    ///
+    /// let a = Atomic::<i32>::null();
+    /// ```
+    pub fn null() -> Self {
```
jeehoonkang: As @Vtec234 pointed out, it would be nice to implement `pub const fn null() -> Self` for nightly.

src/atomic.rs
```diff
@@ -0,0 +1,860 @@
...
+        match self.data.compare_exchange_weak(current.data, new.data, ord, fail_ord) {
+            Ok(_) => {
+                let data = new.data;
+                mem::forget(new);
+                Ok(Ptr::from_data(data))
+            }
+            Err(previous) => Err((Ptr::from_data(previous), new)),
+        }
+    }
+}
```
jeehoonkang: As @Vtec234 pointed out, it would be nice to implement `fetch_??` series (as defined in stdlib).

src/atomic.rs
```diff
@@ -0,0 +1,860 @@
...
+    /// ```
+    pub fn compare_and_swap<'scope>(
+        &self,
+        current: Ptr<T>,
+        new: Ptr<T>,
+        ord: Ordering,
+        _: &'scope Scope,
+    ) -> Result<(), Ptr<'scope, T>> {
+        let fail_ord = strongest_failure_ordering(ord);
+        match self.data.compare_exchange(current.data, new.data, ord, fail_ord) {
```
jeehoonkang: I feel it would be more convenient to just use `self.data.compare_and_swap` instead.  What do you think?

jeehoonkang (changes requested):

src/atomic.rs
ghost: Done

src/atomic.rs
ghost: IMHO it's about the same (`compare_exchange` is more verbose, but on the other hand it allows using `match` on a `Result`, which I prefer to if-statements). Anyways, I changed to `compare_and_swap`.

src/atomic.rs
ghost: Done.

src/atomic.rs
ghost: Personally, I tend to avoid `compare_exchange` because:

1. It's so verbose.
2. `compare_and_swap` with `strongest_failure_ordering` usually does exactly what I want anyway.
3. I'm slightly afraid of screwing up the two orderings (choosing one ordering is difficult enough already :)).

We could add 4 new methods: `compare_exchange{,_weak,_owned,_weak_owned}`, although I'd prefer not adding them right now. But if someone really needs them in the future, I won't object to their addition. How does that sound? :)

src/atomic.rs
```diff
@@ -242,10 +266,11 @@ impl<T> Atomic<T> {
         ord: Ordering,
         _: &'scope Scope,
     ) -> Result<(), Ptr<'scope, T>> {
-        let fail_ord = strongest_failure_ordering(ord);
-        match self.data.compare_exchange(current.data, new.data, ord, fail_ord) {
-            Ok(_) => Ok(()),
-            Err(previous) => Err(Ptr::from_data(previous)),
+        let previous = self.data.compare_and_swap(current.data, new.data, ord);
+        if previous == current.data {
```
Vtec234: Because of the extra comparison, this change adds an extra `cmp` on x86, see https://godbolt.org/g/5uo8AB . Not that it actually matters (I benchmarked it and there's no difference), just FYI.

ghost:
One more thing: function `Ptr::from_raw` is currently marked as unsafe, but should it be?

Before we answer this question, let's consider two related examples:
1. `Box::from_raw` is unsafe because a constructed box can be safely dereferenced.
2. `AtomicPtr::from_raw` is safe because dereferencing is unsafe.

`Ptr` cannot be safely dereferenced. The current contract of `Ptr::as_ref` is:

* It is safe if correct orderings/fences were used.
* Since `Ptr::from_raw` is unsafe, it has already promised that the pointer is valid.

Therefore, the only thing a caller of `Ptr::from_raw` must promise is correct synchronization.

I wonder if all these safety shenanigans are too complicated. :)
To make matter simpler, we could do the following:

* Mark `Ptr::from_raw` as safe.
* Say that `Ptr::as_ref` is safe only if properly synchronized *and* if the pointer is valid.

What do you think?

src/atomic.rs
ghost: Interesting. Thanks for investigating this!

It seems that `compare_exchange` is strictly better here, and I'd rather revert back to it. @jeehoonkang, do you agree?

src/atomic.rs
schets: I'm not a big fan of method bloat but I also don't like the idea of having a bunch of extra memory fences. It's probably not that big of a deal since most of the users will be on x86, but it's worth thinking about.

I don't think that the verbosity is much of a problem in writing compare_exchange, and I'd be fine with removing compare_and_swap completely in favor of compare_exchange.

src/atomic.rs
jeehoonkang: I agree with @stjepang.  This fact enlightens me :) and now I have a different opinion on the methods.  Will leave a comment below.

jeehoonkang:
I agree with @stjepang for marking `Ptr::from_raw` safe.  `Ptr` is more like `AtomicPtr` than `Box` in my opinion.

src/atomic.rs
jeehoonkang: I am suggesting these API changes:

- Support only `compare_exchange*` and deprecate `compare_and_swap*`.  They are too similar to each other and we want to avoid method bloat.

    + Maybe we can rename `compare_exchange` into `compare_and_swap`.  I don't have a strong opinion on this, though.

- Change the signature of `compare_exchange*`.  Now it accepts `orders: Into<Orderings>` rather than `success: Ordering, failure: Ordering`.  Orderings is effectively a pair of success & failure orderings:

            pub fn compare_exchange<'a>(&self, old: Option<Shared<T>>, new: Option<Shared<T>>,
                                            orders: Into<Orderings>, guard: &'a Guard)
                                            -> Result<Option<Shared<'a, T>>, Option<Shared<'a, T>>>

- Implement an implicit conversion from Ordering to Orderings using `strongest_failure_ordering`.

I believe in this way, we can avoid method bloat, and support an ergonomic & generic API at the same time.  What do you think?

arthurprs:
It's probably ok to make it safe, dereferencing is unsafe anyway.

src/atomic.rs
ghost: Regarding naming, here's my breakdown of possible names:

1. `compare_and_swap` - the "swap" here is misleading because the returned value on success is not the old one. That really isn't a swap. :)
2. `compare_exchange` - here, "exchange" has equivalent meaning to "swap". Again, it doesn't exchange values on success. Moreover, the behavior of `compare_exchange` in `libstd` is different.
3. `compare_and_set` - doesn't have any misleading connotations, as far as I can tell. Perhaps this is the most correct naming choice?

I like the `Into<Orderings>` idea, seems like a good solution. :) I've implemented something like it in the form of `CasOrdering` trait. Let me know what you think about that.

jeehoonkang: - I like the implementation :)  I just prefer `CASOrdering` to `CasOrdering`, as CAS is already an acronym.

- for the name of the function, I suggest:
    + let the function return `Ok(previous)` when `previous` is the previous value.
    + let's name it `compare_exchange`.

  rationale: this proposal provides the most general interface for CAS-like atomic updates.  In the case that we only need a degenerated interface, the compiler should optimize out the costs incurred by the additional functionality.  In that case, `compare_exchange` seems the natural name for it.

  More thoughts: personally I think that returning `Ok(())` or `Ok(previous)` when successful, is not an important difference.  Information-wise, these two are exactly the same.  Also I think this function is actually a swap, since if the swap was successful, the caller realizes that `old` is actually the old value.  It seems I am not satisfied with `compare_and_set` because it usually means something weaker than `compare_and_swap`: returning only boolean values (successful or not).

  I tried to be as clear as possible on this issue.  I trust @stjepang 's best judgment., so please discuss with other members and decide on it.

Thank you for your efforts, and sorry for late responses!

jeehoonkang (approved):

src/atomic.rs
ghost: Fair enough, let's go with `compare_exchange` then. :)
I've renamed `CasOrdering` to `CompareExchangeOrdering`. It's a bit lengthy, but the meaning is clear.

About returning `Ok(previous)`... My main gripe is with the return type of `compare_exchange_owned`. At the moment it returns:
```rust
Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)>
```

If we change the signature so that it returns the previous value in the success case as well, the return type would have to be:
```rust
Result<(Ptr<'scope, T>, Ptr<'scope, T>), (Ptr<'scope, T>, Owned<T>)>
```
This is a bit long and possibly error-prone because nothing distinguishes the two pointers in the success case (the new and the previous pointer).

For that reason, I think it's best not to return the previous value in the success case.
But now the danger is that someone might take a look at these signatures and think that they follow the same conventions:
```rust
AtomicPtr::compare_exchange(...) -> Result<*mut T, *mut T>
Atomic::compare_exchange_owned(...) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)>
```

But they don't! :) The first function returns the *previous* value in the success case, while the second returns the *new* value. I'm afraid that this might cause some confusion, but perhaps I'm wrong.

ghost:
Here's a poll for some bikeshedding. :)

**Question**: What should be the names and return types of `compare_exchange` and `compare_exchange_owned`?

Breakdown of possible options and their pros/cons follows.

#### First option

Vote this comment with `+1` emoji.

```rust
// Returns () on success.
Atomic::compare_exchange(...) -> Result<(), Ptr<'scope, T>>`

// Returns the new pointer on success.
Atomic::compare_exchange_owned(...) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)>`
```

Pro: They look a lot like the established `Atomic::compare_exchange`.

Con: Consider these two signatures:
```rust
AtomicPtr::compare_exchange(...) -> Result<*mut T, *mut T>
Atomic::compare_exchange_owned(...) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)>
```

You might think that these two follow the same conventions, but they don't!
* `AtomicPtr::compare_exchange` returns the *previous* poiner on success.
* `Atomic::compare_exchange_owned` returns the *new* poiner on success.

That might be confusing.

#### Second option

Vote this comment with `-1` emoji.

```rust
// Returns the old pointer on success.
Atomic::compare_exchange(...) -> Result<Ptr<'scope, T>, Ptr<'scope, T>>

// Returns the old pointer and the new pointer on success.
Atomic::compare_exchange_owned(...) -> Result<(Ptr<'scope, T>, Ptr<'scope, T>), (Ptr<'scope, T>, Owned<T>)>
```

Pro: Doesn't suffer from the same problem the first option does.

Con: The owned variant is very long and the two pointers in the success case are easy to mix up.

#### Third option

Vote this comment with `:D` emoji.

```rust
// Returns () on success.
Atomic::compare_and_set(...) -> Result<(), Ptr<'scope, T>>

// Returns the new pointer on success.
Atomic::compare_and_set_owned(...) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)>
```

Pro: Users are less likely to incorrectly assume that these behave exactly the same as `Atomic::compare_exchange`.

Con: Breaks established conventions: compare-and-set typically doesn't return a `Result`, but just a `bool`.

#### Fourth option

Vote this comment with "hooray" emoji.

```rust
// Returns () on success.
Atomic::compare_and_swap(...) -> Result<(), Ptr<'scope, T>>

// Returns the new pointer on success.
Atomic::compare_and_swap_owned(...) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)>
```

Pro: Users are less likely to incorrectly assume that these behave exactly the same as `Atomic::compare_exchange`.

Con: Breaks established conventions: compare-and-swap typically doesn't return a `Result`, but just the previous pointer (similar to `AtomicPtr::swap` - the new pointer is stored and the previous one is returned).

EDIT: Another con of the fourth option, I think, is that the "swap" part in "compare-and-swap" might be misleading. It indicates that the pointer gets swapped, i.e. that in case of success the new pointer is stored and the previous one is returned (that's what "swap" is). But that is *not* the behavior of the function: on success the new pointer is returned!

jeehoonkang:
@stjepang Your argument on the cons on the first option (and the second one) is very convincing.  I drop my support for it.  Thank you for the analysis.

Among `compare_and_set` and `compare_and_swap`, I prefer the latter since it sounds stronger (in the sense of capability) than `compare_and_set` to my ears.   @stjepang , please wait for another day or two for others and decide on it.  I think it's better to move forward even if there is a (minor) disputed issue.

PR #2, opened 2017-07-05
Implement a participants registry

ghost:
This PR implements thread registration using a lock-free linked list of participants and makes sure that old entries are properly reclaimed (note that reclamation itself is not implemented yet).

The implementation is mostly based on the code in Crossbeam's main repository and Coco. As previously discussed, it might be good idea to replace the linked list with a more cache-efficient data structure. I wrote a "FIXME" comment so that we don't forget to experiment with that.

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -786,7 +786,7 @@ impl<T> AsMut<T> for Owned<T> {
 ///
 /// The pointer must be properly aligned. Since it is aligned, a tag can be stored into the unused
 /// least significant bits of the address.
-#[derive(Copy, Debug)]
+#[derive(Debug)]
 pub struct Ptr<'scope, T: 'scope> {
     data: usize,
     _marker: PhantomData<&'scope T>,
@@ -800,6 +800,8 @@ impl<'scope, T> Clone for Ptr<'scope, T> {
         }
     }
 }
+
+impl<'scope, T> Copy for Ptr<'scope, T> {}
 
 impl<'scope, T> Ptr<'scope, T> {
     /// Returns a new pointer pointing to the tagged pointer `data`.
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -1,54 +1,12 @@
 use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT};
-use std::sync::atomic::Ordering::{AcqRel, Acquire, Relaxed, Release};
 
-use {Atomic, Owned, Ptr, Scope, unprotected};
+use unprotected;
 use participant::Participant;
+use sync::list::{List, Entry};
 
 // FIXME(stjepang): Participants are stored in a linked list because linked lists are fairly easy
 // to implement in a lock-free manner. However, traversal is rather slow due to cache misses and
 // data dependencies. We should experiment with other data structures as well.
-
-/// An entry in the linked list of participating threads.
-struct Entry {
-    /// State of the participating thread.
-    participant: Participant,
-
-    /// The next entry in the linked list.
-    /// If the tag is 1, this entry is marked as deleted.
-    next: Atomic<Entry>,
-}
-
-/// Registers a thread by inserting a new entry into the list of participating threads.
-///
-/// Returns a pointer to the newly allocated and inserted entry.
-fn register() -> *const Entry {
-    let list = participants();
-
-    let mut new = Owned::new(Entry {
-        participant: Participant::new(),
-        next: Atomic::null(),
-    });
-
-    unsafe {
-        // Since we don't dereference any pointers in this block, it's okay to use `unprotected`.
-        unprotected(|scope| {
-            let mut head = list.load(Relaxed, scope);
-
-            loop {
-                new.next.store(head, Relaxed);
-
-                // Try installing the new entry as the new head.
-                match list.compare_and_set_weak_owned(head, new, Release, scope) {
-                    Ok(n) => return n.as_raw(),
-                    Err((h, n)) => {
-                        head = h;
-                        new = n;
-                    }
-                }
-            }
-        })
-    }
-}
 
 thread_local! {
     /// The thread registration harness.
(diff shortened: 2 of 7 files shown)


src/atomic.rs
```diff
@@ -801,6 +801,8 @@ impl<'scope, T> Clone for Ptr<'scope, T> {
     }
 }
 
+impl<'scope, T> Copy for Ptr<'scope, T> {}
```
Vtec234, line 804: I'm curious, what's the difference?

ghost: I'm curious as well, actually... :/

Both `Copy` and `Clone` could be simply derived here. But if any of those are derived, then I get errors like:

```
error[E0507]: cannot move out of borrowed content
   --> src/registry.rs:139:38
    |
139 |                     .compare_and_set(self.curr, succ, AcqRel, self.scope)
    |                                      ^^^^ cannot move out of borrowed content
```

arthurprs: Strange indeed, I assumed the derive generated code would be exactly that.

jeehoonkang: indeed very strange..  how about leaving a "FIXME" for now?  I think deriving `Copy` is much more idiomatic.

src/registry.rs
```diff
@@ -0,0 +1,164 @@
...
+    unsafe {
+        // Since we don't dereference any pointers in this block, it's okay to use `unprotected`.
+        unprotected(|scope| {
+            let mut head = list.load(Acquire, scope);
+
+            loop {
+                new.next.store(head, Relaxed);
+
+                // Try installing the new entry as the new head.
+                match list.compare_and_set_weak_owned(head, new, AcqRel, scope) {
```
jeehoonkang: I think `Release` is okay here, because this thread doesn't need to acquire anything.

src/registry.rs
```diff
@@ -0,0 +1,164 @@
...
+                }
+            }
+        })
+    }
+}
+
+thread_local! {
+    /// The thread registration harness.
+    ///
+    /// The harness is lazily initialized on its first use, thus registrating the current thread.
```
jeehoonkang, line 14: I am just curious on Rust static's semantics: Is this "lazily" initialized?

src/registry.rs
```diff
@@ -0,0 +1,164 @@
...
+
+            if succ.tag() == 1 {
+                // This thread has exited. Try unlinking it from the list.
+                let succ = succ.with_tag(0);
+
+                if self.pred
+                    .compare_and_set(self.curr, succ, AcqRel, self.scope)
+                    .is_err()
+                {
+                    // We lost the race to unlink this entry. There's no option other than to
```
jeehoonkang: What about rereading `self.pred` once again?

src/registry.rs
```diff
@@ -0,0 +1,164 @@
...
+/// Note that the iterator might return the same participant multiple times.
+pub fn iter(scope: &Scope) -> Iter {
+    let pred = participants();
+    let curr = pred.load(Acquire, scope);
+    Iter { scope, pred, curr }
+}
+
+/// An iterator over all participating threads.
+///
+/// Note that the iterator might return the same participant multiple times.
```
jeehoonkang: As far as I understood, this iterator behaves like RCU linked lists, in the sense that insertion, removal, and traversal are not linearizable.  For example, `Harness:drop()` is performing `fetch_or` on `Entry::next` with `Release` ordering, but a concurrent traversal may ignore it.

I think it would be nice if we could comment more on this data structure.  What is the specification?  and Why it works?

src/registry.rs
ghost: Yes, according to the docs:
https://doc.rust-lang.org/std/thread/struct.LocalKey.html#initialization-and-destruction

src/atomic.rs
ghost: Oh, I think I know what's up. Derived `Clone` and `Copy` impls require `T: Clone`.
The error occurs because there is no implementation of `Copy` since `Entry: !Clone`.

src/registry.rs
ghost: You're right. For the same reason it's okay to use `Relaxed` in line 38.

src/registry.rs
ghost: I think we could do simply reread it. Then, if the freshly read pointer is tagged with 1 as well, we must restart from the beginning. Not sure if that's worth the additional several lines of code, though...

src/registry.rs
ghost: I've added a more elaborate comment. Let me know if something is still missing.

jeehoonkang:
@stjepang In my branch, I refactored the code a little bit: (1) I defined the linked list data structure, and (2) let `registry.rs` use it.  What do you think?  Please have a look: https://github.com/jeehoonkang/crossbeam-epoch/commit/02b90293aa48a22fc974f4e4d48e0d3be2ab1ed6

jeehoonkang:
@stjepang I am sorry -- I mistakenly pushed to your branch, not my branch.  My bad.  Please ignore my commit on your branch.

jeehoonkang:
I implemented RFC 3 "epoch gc" in: https://github.com/jeehoonkang/crossbeam-epoch/tree/epoch-gc

I am not making a PR right now because it is (1) not documented, (2) not tested, and (3) maybe too big.  I will do it soon.  I wonder if you guys can have a look in the meantime.

@stjepang I copy-and-pasted many of Coco's code, so maybe it is a copyright infringement.  I would like to get your endorsement that it is okay, if it is indeed okay for you.  Of course I will credit your work in PRs and commits after splitting this big one.

ghost:
@jeehoonkang Thank you, the refactoring looks pretty good! I've merged it into this PR.

Regarding Coco, copyright, and crediting, just don't worry about it. :) Feel free to copy the code as you wish.

jeehoonkang (approved):

PR #3, opened 2017-07-25
[WIP] Epoch gc

jeehoonkang:
WARNING: this is not meant to be merged.  It needs more work.  Comments will be very much appreciated!


jeehoonkang:
Two caveats:

- For the current being, please ignore namespace business (and type variable `N` all around the codebase).  That is code for the scope of epoch management and GC.  There is the one global namespace, which can be used anytime, and a user can create custom namespaces for per-data-structure GC, etc.  I strongly feel that the design of namespace (or even its name!) should walk through an RFC process.  I was just exploring the design space.

- Atomic orderings/instructions are different from the original Crossbeam or Coco.  I think the orderings used in those projects are wrong w.r.t. C/C++ memory model in a very subtle way.  I am preparing an RFC for clarifying this issue.

src/garbage.rs
```diff
@@ -0,0 +1,133 @@
...
+            destroy: unsafe { mem::transmute(destroy) },
+        }
+    }
+
+    /// Make a garbage object that will later be freed.
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `size`
+    /// elements of type `T`.
+    pub fn new_free<T>(object: *mut T, size: usize) -> Self {
+        unsafe fn free<T>(object: *mut T, size: usize) {
```
schets: It would better if this destruct vs not dispatch was changed to be a parameter stored in the Garbage object, since this would reduce the number of distinct indirect function calls.

It would probably be possible to eliminate the indirect call in the case of no destructors and instead free a Vec<u8> of the right size.

It's not a huge difference but prediction hardware for indirect calls is more limited than for branches so I always try to not use it pointlessly

src/registry.rs
```diff
@@ -0,0 +1,62 @@
...
+        // announced to other threads. Only then it becomes safe to load from the shared memory.
+        if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
+            // On x86 architectures we have a choice:
+            // 1. `atomic::fence(SeqCst)`, which compiles to a `mfence` instruction.
+            // 2. `compare_and_swap(_, _, SeqCst)`, which compiles to a `lock cmpxchg` instruction.
+            //
+            // Both instructions have the effect of a full barrier, but the second one seems to be
+            // faster in this particular case.
+            let previous = self.state.load(Relaxed);
+            self.state.compare_and_swap(previous, state, SeqCst);
```
schets: Using an xchg instruction (the swap function) would probably be slightly faster since there's no data dependency on the load from previous. In the case where state wasn't in the cache, which won't show up in microbenchmarks, it would be much faster.

src/registry.rs
```diff
@@ -0,0 +1,62 @@
...
+        let state = self.state.load(Relaxed);
+        ((state & 1) == 1, state & !1)
+    }
+
+    /// Marks the thread as pinned.
+    ///
+    /// Must not be called if the thread is already pinned!
+    #[inline]
+    pub fn set_pinned<N: Namespace>(&self, namespace: N) {
+        let epoch = namespace.epoch().load(Relaxed);
```
schets: It would probably help to remove this load as a data dependency (or even just as a shared load which must be fenced) from pin. If we allow a stale epoch to be read, which I think is safe since it would only slightly delay collection, this might be possible to do efficiently on x86 at least.

It would be hard to benchmark the effect of this this change since microbenchmarks would have this always be in the cache.

src/scope.rs
```diff
@@ -0,0 +1,232 @@
...
+    ///
+    /// [`Atomic`]: struct.Atomic.html
+    pub fn pin<F, R>(&self, f: F) -> R where
+        F: FnOnce(&Scope<N>) -> R,
+    {
+        let registry = self.registry.get();
+        let scope = &Scope { namespace: self.namespace, bag: self.bag.get() };
+
+        let was_pinned = self.is_pinned.get();
+        if !was_pinned {
```
schets: Why not merge is_pinned and pin count, since it looks like is_pinned being true is equivalent to pin_count==0? Unless I'm misreading where the variables are used.

src/registry.rs
jeehoonkang: Honestly I even don't know why it is correct in x86..  Please discuss with @stjepang .  He introduced this optimization in Coco.

src/scope.rs
jeehoonkang: `pin_count` is the number of `pin()` called so far, and it is not related to `is_pinnned`.  It is tracked for occasionally collecting the global garbages (once in `PINS_BETWEEN_COLLECT` times).

src/registry.rs
jeehoonkang: I am not sure I understand your comment correctly, but here is the only place that the global epoch is loaded in `pin()`.

src/scope.rs
schets: ah! I thought pin_count was the depth of the pin stack.

src/garbage.rs
jeehoonkang: Thank you for pointing out that.  I added a commit on it.

src/registry.rs
schets: Currently, pin looks something like this:

```
epoch = load_global_epoch;
state = epoch | 1;
fenced_store_pin_state(epoch); // must wait for the load of epoch to finish
```

where instead a faster way would resemble:

```
// advertise current epoch somehow
fenced_store_pin_state(true);
// possibly afterwards, could result in a stale epoch version being read when pinned
```

but whether that's possible or not is an open question

src/registry.rs
ghost: @schets So if I got this right, you're suggesting that we do the following?
```rust
self.state.swap(state, SeqCst);
```

This will compile to `xchg` rather than `lock xchg`. The latter has the effect of `fence(SeqCst)` (`mfence`), but I'm not sure about the former. Does `xchg` imply a full fence as well?

ghost: On another thought... `previous` will always be 0 here.
So we can just do `self.state.compare_and_swap(0, state, SeqCst)` and now we don't have the data dependency anymore. Problem solved, right?

schets: yeah, xchg is weird and locks whenever it has a memory operand

src/registry.rs
jeehoonkang: @schets As I said before, I am preparing an RFC for clarifying why Crossbeam is correct w.r.t. C/C++ memory model (https://github.com/jeehoonkang/crossbeam-rfcs/blob/relaxed-memory/text/2017-07-23-relaxed-memory.md).  I would like to discuss this issue there, what do you think?

src/registry.rs
jeehoonkang: I addressed @stjepang 's comment.

jeehoonkang:
I addressed all the comments given so far.  Thank you!  I will make a PR based on this branch as follows:

- The code should be documented.
- I will remove the namespaces.  The current premature implementation will likely clutter further discussion.
- I will *not* change the atomic instructions and orderings.  As opposed to namespaces, I think they will bother further discussion.

ghost:
I see a lot of code was pulled from Crossbeam mainline (`CachePadded`, `AtomicOption`, `MsQueue`, scoped threads). If you find that those abstractions don't fit your needs, don't hesitate to implement custom ones. For example, `MsQueue` doesn't provide a `try_pop_if` method, and you will probably need it.

Since this PR is becoming pretty large, feel free to write inefficient (but simple) code and label it with a `FIXME` for a later PR. My suggestion is to implement the epoch-tracking logic now (pinning and advancing epochs), and just leave `defer_free`/`defer_drop`/`defer`/`flush` unimplemented. Then you can implement garbage collection in a follow-up PR.

Finally, I'd encourage using rustfmt. The rustfmt developers would appreciate using it on as many projects as possible. And if you stumble upon any suboptimal formatting, simply report the problem on their issue tracker.

jeehoonkang:
> Since this PR is becoming pretty large, feel free to write inefficient (but simple) code and label it with a FIXME for a later PR. My suggestion is to implement the epoch-tracking logic now (pinning and advancing epochs), and just leave defer_free/defer_drop/defer/flush unimplemented. Then you can implement garbage collection in a follow-up PR.

I will split the PR into one on managing epoch and the other on managing garbage.

> I see a lot of code was pulled from Crossbeam mainline (CachePadded, AtomicOption, MsQueue, scoped threads). If you find that those abstractions don't fit your needs, don't hesitate to implement custom ones. For example, MsQueue doesn't provide a try_pop_if method, and you will probably need it.

I removed blocking `pop()` from MsQueue, and added `try_pop_if()` in it.  The method was needed in `collect()`, which pops a bag from the global queue if the bag's epoch <= current - 2.  MsQueue in turn requires `atomic_option.rs` and `scoped.rs`, which I just copy-and-pasted, for test code.  I also grab `cache_padded.rs` from Crossbeam because I wanted to lay out data in a cacheline.

I think `cache_padded.rs` can be separated into another PR.  In current implementation, `ms_queue.rs`, along with `atomic_option.rs`, `scoped.rs` for test code, is tightly coupled with garbage management, which I intend to make them in the PR on garbage management.  What do you think?

> Finally, I'd encourage using rustfmt. The rustfmt developers would appreciate using it on as many projects as possible. And if you stumble upon any suboptimal formatting, simply report the problem on their issue tracker.

I will do it from now on.  I ran `rustfmt` against the current PR, and added a commit on it.  Thank you for the advice!


I think I cannot read or write in IRC for about two weeks.  For this period, I will continue to work on this PR and read GitHub activities.  Please forgive my absence in IRC ;)

ghost:
>  In current implementation, ms_queue.rs, along with atomic_option.rs, scoped.rs for test code, is tightly coupled with garbage management, which I intend to make them in the PR on garbage management. What do you think?

Sounds good!

I think there are several problems with the following method:

```rust
fn try_pop_if<F>(&self, condition: F, scope: &Scope) -> Option<T>
where
    F: Fn(&T) -> bool,
{
    // ...
}
```

First, this requires an additional type bound `T: Sync`.

Second, suppose that a thread calls `try_pop_if(...)`, which in turn calls `condition` with the current head of the queue. Another thread comes and pops the head from the queue and destroys it. Now the first thread attempts to read the destoyed `T`, and bad things happen. Note that if a thread pops an element from the queue, it *owns* it because the element has been moved out of the queue, and other threads cannot hold a reference to it at the same time.

But if you add one more bound `T: Copy`, this will be totally fine.

> I think I cannot read or write in IRC for about two weeks. For this period, I will continue to work on this PR and read GitHub activities. Please forgive my absence in IRC ;)

No problem :)

ghost:
I took a glance at your relaxed-memory RFC, and it looks pretty good. It's exciting on two fronts: first, it gives us confidence into correctness of the GC, and second, it's very readable documentation of its inner workings.

A question, as a complete dummy with regards to formal verification: how difficult would it be to write a model for our epoch-based GC and formally verify it?

Since the GC is actually rather simple (at least in comparison to data structures built on top of it), I'd expect that to be doable...

jeehoonkang:
> A question, as a complete dummy with regards to formal verification: how difficult would it be to write a model for our epoch-based GC and formally verify it?

Actually I am doing it for my day job as a CS graduate student :)  My current long-term research goal is verifying Crossbeam, Chase-Lev deque, and other concurrent data structures.  I recall that I said relaxed accesses and SC fences are two difficult things for formally verifying Chase-Lev deque.  Actually it is the same for epoch-based reclamation (EBR).

I would like to focus on reasoning SC fences first.  (For research purposes, we can temporarily promote relaxed accesses to acquire/release ones and reason about it.)  Yes, EBR is actually rather simple, but SC fences are still not well understood so far.  I am not aware of a program logic for C11 that properly deals with SC fences.  This is roughly because SC synchronization depends on "negative" reasoning (i.e. the absence of information means something, e.g. I didn't see x=1 so something good is satisfied).  Note that acquire-release synchronization depends on "positive" reasoning (i.e. the presence of information means something, e.g. I did acquire-see release-written x=1, so something good is satisfied).

ghost:
> Actually I am doing it for my day job as a CS graduate student :) My current long-term research goal is verifying Crossbeam, Chase-Lev deque, and other concurrent data structures.

Ohhh, I see. That's fantastic - really looking forward to your research work. :)
Thanks for the clarification!

By the way, yesterday Microsoft Research has published a [paper](https://www.microsoft.com/en-us/research/publication/project-snowflake-non-blocking-safe-manual-memory-management-net/#) on non-blocking manual memory reclamation: Project Snowflake. I've just skimmed over it.

Snowflake enables you to manually allocate objects in C#, which can be concurrently shared, and later they are reclaimed with a combination of hazard pointers and EBR. They use this particular combination because it fits their environment well (a GC-ed and JIT-ed language), but would not be a good fit for us.

Just like in the paper on Deuteronomy/Hekaton, the epoch counter is a 64-bit integer. Threads from time to time load the global epoch and copy it into the TLS - this is akin to pinning. They don't have the 3-step *pin-use-upin* strategy, but instead occasionally update the local epoch, which reminds me a bit of QSBR. This is because the use of concurrently shared objects is always pervasive rather than tightly scoped between the *pin* and *upin* steps.

On page 19 they even mention Crossbeam:

> Epoch-based reclamation (EBR) has been used for lock-free data structures [39, 34, 40], and has been very elegantly captured in the Rust Crossbeam library [7].  To access a shared data structure you “pin” the global epoch which prevents it advancing (too much), and when you finish accessing the data structure you release the pin.  As epochs advance the older deallocated objects can be reclaimed.  We could not directly use epochs without hazards, as EBR requires there to be points where you are not accessing the protected data structure.  When EBR is protecting a particularly highly-optimised data structure, this is perfectly sensible. We allow pervasive use of manually allocated objects and thus enforcing points where the manual heap is not being accessed is simply impractical.  One could view shields as pinning a single object, rather than a global pin as in Crossbeam.

I don't think there's anything in the paper directly applicable or useful to us, but it's definitely an interesting read. :)

They've also found a way to avoid unbounded garbage growth when a thread falls behind - there's a way to *eject* a stale thread, i.e. temporarily evict it from the EBR scheme so that garbage collection can move on. When the thread catches up, it must reregister. They use a few OS tricks (`VirtualProtect` in Windows, and `mprotect` in Linux) to achieve that. It's all fairly involved and complicated, but pretty cool.

ghost:
Should we close this PR?

PR #4, opened 2017-07-30
Implement epoch management

jeehoonkang:
This is a subset of #3.  It implements epoch management.  The garbage management part will make a separate PR.


jeehoonkang:
The test will fail because some methods are `unimplemented!()` yet.  In order to ensure the correctness of this implementation, I rebased #3 based on this commit, which passes the test.

src/atomic.rs
```diff
@@ -216,7 +216,8 @@ impl<T> Atomic<T> {
     ///     let p = a.load(SeqCst, scope);
     /// });
     /// ```
-    pub fn load<'scope>(&self, ord: Ordering, _: &'scope Scope) -> Ptr<'scope, T> {
+    pub fn load<'scope>(&self, ord: Ordering, _: &'scope Scope) -> Ptr<'scope, T>
+where {
```
ghost: A bunch of redundant `where`s in this file...

src/epoch.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Returns the current global epoch.
+    #[cold]
+    pub fn try_advance<'scope>(&self, registries: &List<Registry>, scope: &Scope) -> usize {
+        let epoch = self.epoch.load(Relaxed);
+        ::std::sync::atomic::fence(SeqCst);
+
+        // Traverse the linked list of mutator registries.
+        let mut registries = registries.iter(scope);
+        loop {
+            match registries.next() {
```
ghost, line 44: I like the way iteration works here. :)

src/epoch.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Attempts to advance the global epoch.
+    ///
+    /// The global epoch can advance only if all currently pinned mutators have been pinned in the
+    /// current epoch.
+    ///
+    /// Returns the current global epoch.
+    #[cold]
+    pub fn try_advance<'scope>(&self, registries: &List<Registry>, scope: &Scope) -> usize {
+        let epoch = self.epoch.load(Relaxed);
+        ::std::sync::atomic::fence(SeqCst);
```
ghost: Why don't we just have a SC load instead of a SC fence?
Are fences here even necessary? If possible, I'd like to avoid fences because thread sanitizer doesn't recognize them, which might become a nuisance later on.

src/epoch.rs
```diff
@@ -0,0 +1,82 @@
...
+
+                    // If the mutator was pinned in a different epoch, we cannot advance the global
+                    // epoch just yet.
+                    if mutator_is_pinned && mutator_epoch != epoch {
+                        return epoch;
+                    }
+                }
+            }
+        }
+        ::std::sync::atomic::fence(Acquire);
```
ghost, line 62: Why this fence?

src/mutator.rs
```diff
@@ -0,0 +1,287 @@
...
+
+
+/// Number of pinnings after which a mutator will collect some global garbage.
+const PINS_BETWEEN_COLLECT: usize = 128;
+
+
+/// Entity that changes shared locations.
+pub struct Mutator<'scope> {
+    /// This mutator's entry in the registry list.
+    registry: &'scope Node<Registry>,
```
ghost: Why not `&'static Node<Registry>`? Will it ever make sense to have a shorter lifetime?

src/sync/list.rs
```diff
@@ -1,125 +1,175 @@
 use std::sync::atomic::Ordering::{Acquire, Relaxed, Release};
 
-use {Atomic, Owned, Ptr, Scope};
+use {Atomic, Owned, Ptr, Scope, unprotected};
+use util::cache_padded::CachePadded;
+
 
 /// An entry in the linked list.
-pub struct Entry<T> {
+struct Node_<T> {
```
ghost: Why a trailing underscore?

src/sync/list.rs
```diff
@@ -1,125 +1,175 @@
...
             }
         }
     }
 
+    pub fn insert_head<'scope>(
+        &'scope self,
+        data: T,
+        scope: &'scope Scope,
+    ) -> Ptr<'scope, Node<T>>
+where {
```
ghost: Unncessary where

src/sync/list.rs
```diff
@@ -1,125 +1,175 @@
...
     ///
     /// 1. If a new datum is inserted during iteration, it may or may not be returned.
     /// 2. If a datum is deleted during iteration, it may or may not be returned.
-    /// 3. Any datum that gets returned may be returned multiple times.
-    pub fn iter<'scope>(&'scope self, scope: &'scope Scope) -> Iter<'scope, T> {
-        let head = &self.head;
-        let pred = head;
+    /// 3. It may not return all data if a concurrent thread continues to iterate the same list.
+    pub fn iter<'scope>(&'scope self, scope: &'scope Scope) -> Iter<'scope, T>
+where {
```
ghost: `where`

src/sync/list.rs
```diff
@@ -1,125 +1,175 @@
...
-impl<'scope, T> Iterator for Iter<'scope, T> {
-    type Item = &'scope T;
+impl<T> Drop for List<T> {
+    fn drop(&mut self) {
+        unsafe {
+            unprotected(|scope| {
+                let mut curr = self.head.load(Relaxed, scope);
+                while let Some(c) = curr.as_ref() {
+                    let succ = c.0.next.load(Relaxed, scope);
+                    scope.defer_free(curr);
```
ghost: In destructors you don't have to defer destruction - you can destroy immediately.

ghost (changes requested):
A few minor comments.

This is not a good place for `MsQueue` and scoped threads, but I'm okay with keeping them temporarily until we refactor the `crossbeam` crate.

src/util/lazy_static.rs
```diff
@@ -0,0 +1,62 @@
...
+        lazy_static!($VISIBILITY, $NAME, $T, Default::default());
+    };
+}
+
+/// $NAME() returns a reference to the static object, which is lazily initialized to null.
+#[macro_export]
+macro_rules! lazy_static_null {
+    ($VISIBILITY:tt, $NAME:ident, $T:ty) => {
+        #[inline]
+        $VISIBILITY fn $NAME() -> &'static $T {
```
Vtec234: How about defining `get()` as in `lazy_static` instead? It would be a bit clearer than the function-object-like syntax `variable()`.

src/lib.rs
```diff
@@ -1,10 +1,48 @@
...
+//!
+//! # Pointers
+//!
+//! Concurrent collections are built using atomic pointers. This module provides [`Atomic`], which
+//! is just a shared atomic pointer to a heap-allocated object. Loading an [`Atomic`] yields a
+//! [`Ptr`], which is an epoch-protected pointer through which the loaded object can be safely read.
+//!
+//! # Pinning
+//!
+//! Before an [`Atomic`] can be loaded, the current mutator must be [`pin`]ned. By pinning a mutator
```
Vtec234, line 11: Is there any advantage in using terminology of "mutators" over "threads"? Can anything other than a thread be a mutator? In any case, it would be good to at least clarify it by changing "current mutator" to "current mutator (current thread)".

And if indeed only threads can be mutators, maybe `MutatorThread` or such would be a clearer name.

src/lib.rs
```diff
@@ -1,10 +1,48 @@
...
+//!
+//! # Garbage
+//!
+//! Objects that get removed from concurrent collections must be stashed away until all currently
+//! pinned mutators get unpinned. Such objects can be stored into a [`Garbage`], where they are kept
+//! until the right time for their destruction comes.
+//!
+//! There is a global shared instance of garbage queue, which can only deallocate memory. It cannot
+//! drop objects or run arbitrary destruction procedures. Removed objects can be stored into it by
+//! calling [`defer_free`] or [`defer_drop`]. Also you can [`defer`] to call an arbitrary function
```
Vtec234: If the global garbage queue cannot drop objects, why can we add objects to it with `defer_drop`? Same for arbitrary functions and `defer`.

src/sync/list.rs
Vtec234: Please correct me if I'm wrong. Since this structure is used to facilitate the `Registry` list, _can_ we even use `defer_free` here? This will be dropped at the end of program execution when the statics are destroyed, but presumably `defer_free` will refer to the global participants list, which is the thing currently being destroyed.

src/sync/ms_queue.rs
```diff
@@ -0,0 +1,417 @@
...
+
+/// A Michael-Scott lock-free queue, with support for blocking `pop`s.
+///
+/// Usable with any number of producers and consumers.
+// The representation here is a singly-linked list, with a sentinel
+// node at the front. In general the `tail` pointer may lag behind the
+// actual tail. Non-sentinel nodes are either all `Data` or all
+// `Blocked` (requests for data from blocked threads).
+#[derive(Debug)]
+pub struct MsQueue<T> {
```
Vtec234: As @stjepang noted, this is not a very good place to include `MsQueue` and scoped threads. Since they are not used anywhere in the epoch code, maybe we could merge them later in a separate PR and focus on the GC in this one?

src/mutator.rs
```diff
@@ -0,0 +1,287 @@
...
+        // Clear the last bit.
+        // We don't need to preserve the epoch, so just store the number zero.
+        self.state.store(0, Release);
+    }
+}
+
+impl Scope {
+    /// Deferred deallocation of heap-allocated object `ptr`.
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `count`
```
Vtec234: Seems like a leftover comment from a function with a different signature.

src/mutator.rs
```diff
@@ -0,0 +1,287 @@
...
+    ///
+    /// [`Bag`]: struct.Bag.html
+    /// [`flush`]: fn.flush.html
+    pub unsafe fn defer_free<T>(&self, ptr: Ptr<T>) {
+        unimplemented!()
+    }
+
+    /// Deferred destruction and deallocation of heap-allocated object `ptr`.
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `count`
```
Vtec234: Same here.

src/sync/mod.rs
```diff
@@ -1 +1,4 @@
+//! Synchronization primitives.
+
 pub mod list;
```
Vtec234, line 3: Do we want to expose this `List` implementation to the outside world? Imo we should keep it as a detail of the epoch GC since e.g. the `Iter` requires `Scope`s, but as I understand scopes are only used when implementing data structures. When using them, there should be a simpler API.

Vtec234 (changes requested):
I'm afraid I do not have the expertise to find subtle issues if such exist, but at least from a high-level this looks good. Left some minor comments, most importantly the "mutator" vs "thread" naming and whether we should keep `MsQueue` here.

src/epoch.rs
jeehoonkang: @stjepang @Vtec234 I think SC reads/writes are not sufficient to achieve the required synchronization.  For example, Crossbeam needs such synchronization:

```
clode block X;
read global epoch G
||
read global epoch G+1
code block Y;     // should know of the code block X, as if release/acquire synchronization is done
```

According to a recent understanding of SC semantics (http://plv.mpi-sws.org/scfix/), SC reads/writes (of G) cannot provide such guarantee, and we should use SC fence's "cumulativity" for that purposes.

I am clarifying this point in the upcoming RFC (https://github.com/jeehoonkang/crossbeam-rfcs/blob/relaxed-memory/text/2017-07-23-relaxed-memory.md).  Shall we discuss the orderings and correctness in that RFC?

src/sync/list.rs
jeehoonkang: I `CachePadded` this `Node_` to make `Node`.  I also don't like this name, but couldn't find a better one.  Would you please suggest one?

src/sync/list.rs
jeehoonkang: @stjepang yes, that's a good point.  I will revise it.

@Vtec234 that's a very good question, and probably I should have written the reason in the comment.  In fact, `registries()` is a static `AtomicUsize`, and the destructor is not called at all, even at program exit.  So I think it is safe to `defer_free()`, while it is sub-optimal as @stjepang pointed out.

src/util/lazy_static.rs
jeehoonkang: At first, I hand-wrote this macros because I heard `lazy_static` crate is not lock-free (https://github.com/crossbeam-rs/crossbeam/pull/122#issuecomment-288436809).  But these hand-written macros are really sub-optimal.

I would rather use `lazy_static` for the current PR, and later we can optimize it with a better implementation.  What do you think?

src/lib.rs
jeehoonkang: Thank you for the good catch.  I updated the documentation.

src/sync/ms_queue.rs
jeehoonkang: `MsQueue` is used in garbage management.   I will remove it from this PR.

src/sync/mod.rs
jeehoonkang: The `crossbeam_epoch` crate implements epoch-based memory reclamation, but in doing so, it uses concurrent list and queue.  Thus it should implement them.  Then the `crossbeam` crate will depend on `crossbeam_epoch` and implement various data structures.  But list and queue are already implemented in `crossbeam_epoch`, so I thought `crossbeam` can reexport those implemented in `crossbeam_epoch`.  This is the reason I made `list` (and later `ms_queue`) public.  What do you think of this crate structure?

src/lib.rs
jeehoonkang: That is related to @stjepang 's question: will `Mutator` have a lifetime different from `'static`?  (BTW, "mutator" seems an established terminology among garbage collection guys.)  Naming the struct as `Mutator` and allowing other lifetimes than `'static` is an effort for separation of concern.  Files `mutator.rs` and `epoch.rs` concern only with the struct for mutators and the global epoch, respectively.  File `global.rs`'s role is actually defining the global epoch and defining a mutator for each thread.

NB: I came to design after exploring the design space of `Namespace` (now I renamed it to `Realm`).  With that generalization in mind, I needed to separate of the struct (`mutator.rs` and `epoch.rs`, and presumably `garbage.rs` later) and the global instance of realm (`global.rs`).

jeehoonkang:
I addressed all comments from @stjepang and @Vtec234 .  Thank you!

jeehoonkang:
This is the promised RFC on the correctness of Crossbeam: https://github.com/crossbeam-rs/rfcs/pull/6

@schets @stjepang @Vtec234 this RFC is based on what was discussed but postponed in #3 and #4.  Thank you for your inputs!

src/lib.rs
Vtec234: Okay, makes sense. One problem I see is that even though `mutator.rs` is a separate file from `global.rs`, it references the statics defined in there, which makes them intrinsically linked and might make adding other "realms" somewhat complicated.

src/sync/mod.rs
Vtec234: I am not opposed to reexporting structures from internal `crossbeam-rs` crates as part of the public interface, what I mean is rather that this might not be the `List` that we would want to expose. This is a longer discussion that requires its own RFC, but at least my opinion is that the API of this `List` implementation is too low-level. Firstly, it requires the user to provide `Scope` arguments, which the structures currently provided by `crossbeam` don't do. Secondly, the iteration interface has an `Abort` variant. While this is very nice for our purposes, I believe that for an external user who just wants to have a concurrent list, it exposes too much implementation detail and requires said user to understand why exactly iteration can abort, which is a peculiarity of the lock-free implementation.

src/sync/list.rs
Vtec234: Ok, might be good to add a short comment explaining this (it's also guaranteed by the `lazy_static` crate not to run destructors).

src/sync/list.rs
ghost: Such structures are typically called `Inner`, or perhaps `InnerNode`.

src/util/lazy_static.rs
ghost: I'm in favor of using `lazy_static` for now. We can optimize these minor issues later.

src/lib.rs
jeehoonkang: @Vtec234 yes, it is a little bit unfortunate..  In the future, each `Realm` should be able to make a new `Mutator`, and in that case we no longer need to access the global `REGISTRIES` and `EPOCH` from `mutator.rs`, as done in #3.  Note that `mutator.rs` there does not reference the global data.

src/sync/mod.rs
jeehoonkang: I hid `sync` module from the top-level.

ghost: I agree with @Vtec234 here - let's keep it simple and focus on EBR at this point.

`MsQueue`, `TreiberStack`, linked lists, etc. are pretty easy data structures to (re)implement, so I don't see much value in abstracting them away. There will be a time for that later on. :)

src/mutator.rs
```diff
@@ -0,0 +1,282 @@
...
+    registry: &'scope Node<Registry>,
+    /// Whether the mutator is currently pinned.
+    is_pinned: Cell<bool>,
+    /// Total number of pinnings performed.
+    pin_count: Cell<usize>,
+}
+
+/// An entry in the linked list of the registered mutators.
+#[derive(Default, Debug)]
+pub struct Registry {
```
ghost: This is not a registry -- the registry is the linked list that holds thread-local epochs, and this struct is just an entry in the registry. Can we name this struct differently?

jeehoonkang: I renamed it to `LocalEpoch`.  What do you think?

ghost: Much better! :)

src/lib.rs
ghost: @jeehoonkang Could you explain a bit what is the motivation behind *realms*, and in particular why would someone want multiple *realms*?

jeehoonkang: @stjepang I just assumed it was already discussed in the epoch-gc RFC (https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-05-23-epoch-gc.md), but it actually isn't.  Sorry for that.

Each "realm" takes care of the garbages produced in it, but nothing else.  We can allocate a realm for a performance-critical subset of a program so that the code of concern is not penalized with (possibly) a lot of garbages produced elsewhere of the program.  In other words, we can restrict the effect of a bad-behaving neighbor to its own realm only.

We currently have only one global realm, but in the future maybe it is desirable to allow users to create their own realm.  Of course this should go through an RFC process, and I don't mean that we should do it.  I just wanted to explain why I prefer the term "mutator" to "thread" and use a generic lifetime for `Scope`, and why I organized the files (especially `global.rs`) in this way.

Vtec234 (approved):

src/epoch.rs
jeehoonkang: Oh, I forgot to answer this question.  I am sorry.

This is in order to synchronize the release-writes to local epochs (https://github.com/crossbeam-rs/crossbeam-epoch/pull/4/files/42c2fb50e75ceedf71022173b539aa6f75e519d9#diff-1e3fb927a1f124f8fdede26bfa2a730dR211) via the relaxed-read of it (https://github.com/crossbeam-rs/crossbeam-epoch/pull/4/files/42c2fb50e75ceedf71022173b539aa6f75e519d9#diff-6d7fb2f2e4b00caaa3efe3c37b5708e9R52).

jeehoonkang:
I re-read all the discussions so far, and answered & addressed the comments I've overlooked.  Sorry!

ghost (approved):
Let's get this merged and move onto implementing garbage collection.

PR #5, opened 2017-08-14
Implement garbage management

jeehoonkang:
This is a follow-up of #4, and it should not be merged until #4 is merged.  I am creating a PR for sharing the code.

The increment from #4 is https://github.com/crossbeam-rs/crossbeam-epoch/pull/5/commits/bad559163e82672a3ea306eed56a57cd09dea29c.

@stjepang I heard that you have your working implementation (https://github.com/crossbeam-rs/crossbeam/issues/116#issuecomment-321801000).  Can we compare yours and mine in order to make sure that we put best ideas in the merged code?


src/epoch.rs
```diff
@@ -0,0 +1,82 @@
...
+        Self::default()
+    }
+
+    /// Attempts to advance the global epoch.
+    ///
+    /// The global epoch can advance only if all currently pinned mutators have been pinned in the
+    /// current epoch.
+    ///
+    /// Returns the current global epoch.
+    #[cold]
```
arthurprs: Why cold?

ghost:
Can you please rebase on top of master? I've just merged #4.

The implementation I mentioned in that thread was actually `coco`. :) I've just released version 0.2.0 that more closely mirrors the new Crossbeam API so that Tyler can start playing with the new API. But I don't intend to work on `coco` anymore - it'll be deprecated as soon as we finish `crossbeam-epoch`.

jeehoonkang:
I just rebased.  I will soon read coco 0.2.  Thanks!

jeehoonkang:
As I said in #7, I tested this branch (+ #7) against @stjepang's channel.  FYI, here is the benchmark result.  `master` is channel on `coco`, and `epoch` is channel on this branch:

```
bounded0_mpmc             epoch       5.802 sec
bounded0_mpmc             master      6.155 sec

bounded0_mpsc             epoch       5.529 sec
bounded0_mpsc             master      5.886 sec

bounded0_select_both      epoch      65.533 sec
bounded0_select_both      master     65.821 sec

bounded0_select_rx        epoch       6.293 sec
bounded0_select_rx        master      6.607 sec

bounded0_spsc             epoch       5.981 sec
bounded0_spsc             master      5.868 sec

bounded1_mpmc             epoch       5.190 sec
bounded1_mpmc             master      5.004 sec

bounded1_mpsc             epoch       4.132 sec
bounded1_mpsc             master      3.644 sec

bounded1_select_both      epoch       2.992 sec
bounded1_select_both      master      2.893 sec

bounded1_select_rx        epoch       3.572 sec
bounded1_select_rx        master      4.227 sec

bounded1_spsc             epoch       1.706 sec
bounded1_spsc             master      1.580 sec

bounded_mpmc              epoch       2.526 sec
bounded_mpmc              master      2.599 sec

bounded_mpsc              epoch       2.171 sec
bounded_mpsc              master      2.017 sec

bounded_select_both       epoch       0.921 sec
bounded_select_both       master      0.880 sec

bounded_select_rx         epoch       0.936 sec
bounded_select_rx         master      0.932 sec

bounded_seq               epoch       0.267 sec
bounded_seq               master      0.283 sec

bounded_spsc              epoch       0.194 sec
bounded_spsc              master      0.191 sec

unbounded_mpmc            epoch       2.067 sec
unbounded_mpmc            master      2.089 sec

unbounded_mpsc            epoch       1.491 sec
unbounded_mpsc            master      1.415 sec

unbounded_select_both     epoch       0.963 sec
unbounded_select_both     master      0.913 sec

unbounded_select_rx       epoch       1.273 sec
unbounded_select_rx       master      1.070 sec

unbounded_seq             epoch       1.040 sec
unbounded_seq             master      0.564 sec

unbounded_spsc            epoch       1.374 sec
unbounded_spsc            master      1.230 sec
```

Overall, I feel this branch is comparably **slower** than `coco`.  Though I am entirely not sure of, I suspect it is due to the use of `lazy_static!` in this branch.  I read `coco` 0.2 thoroughly, but I couldn't imagine other explanation of this performance difference.. but we will see.  I will investigate further on the performance issue.

ghost:
Well, the only benchmarks that actually touch coco/crossbeam-epoch are `unbounded_mpmc` and `
unbounded_select_both`. So I think there is a lot of noise in those benchmarks. :)

jeehoonkang:
@stjepang Thank you for clarification.  Then maybe we can say that `coco` and `crossbeam-epoch` performs comparably for `channel`.  But it makes me more puzzled, as other test cases reveal unexplicable performance difference..

ghost:
Yeah, they do perform comparably. Thanks for doing the benchmarks. I'll look more into discrepancies in tests like `unbounded_seq` - that one is really weird.

Vtec234:
FWIW, I ran the two benchmarks and `unbounded_select_both` is slower with `epoch` by `.173 sec` on average. `unbounded_mpmc` is slower by `.031 sec`, but that's probably not significant. This should be investigated further, hopefully the other results are just noise.

src/mutator.rs
```diff
@@ -213,8 +226,24 @@ impl LocalEpoch {
...
+
+        while let Err(g) = bag.try_push(garbage) {
+            global::push_bag(bag, self);
+            garbage = g;
+        }
+    }
+
     /// Deferred deallocation of heap-allocated object `ptr`.
     ///
+    /// The specified object is an array allocated at address `object` and consists of `count`
```
Vtec234: Heh, this comment again. It does not belong here :) Same on `defer_drop`

src/garbage.rs
```diff
@@ -0,0 +1,140 @@
+//! Garbage collection.
+//!
+//! # Garbages
+//!
+//! Objects that get unlinked from concurrent data structures must be stashed away until the global
+//! epoch sufficiently advances so that they become safe for destruction.  We call these object
```
Vtec234: "objects"

src/garbage.rs
```diff
@@ -0,0 +1,140 @@
...
+//! is marked with the current global epoch and pushed into a global queue of garbage bags.  We
+//! store garbages in thread-local storages for amortizing the synchronization cost of pushing the
+//! garbages to a global queue.
+//!
+//! # Garbage queues
+//!
+//! Whenever a bag is pushed into a queue, some garbage in the queue is collected and destroyed
+//! along the way.  Garbage collection can also be manually triggered by calling `collect()`.  This
+//! design reduces contention on data structures.  Ideally each instance of concurrent data
+//! structure may have it's own queue that gets fully destroyed as soon as the data structure gets
```
Vtec234: "its" (sorry)

src/garbage.rs
```diff
@@ -0,0 +1,140 @@
...
+
+unsafe impl Send for Garbage {}
+
+impl Garbage {
+    /// Make a garbage object that will later be destroyed using `destroy`.
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `size`
+    /// elements of type `T`.
+    ///
+    /// Note: The object must be `Send + 'self`.
```
Vtec234: Hm, I think I get what you mean here by `'self`, but technically this isn't valid Rust. Maybe use `'static` instead to reduce confusion?

src/garbage.rs
```diff
@@ -0,0 +1,140 @@
...
+            size: mem::size_of::<T>() * size,
+        }
+    }
+
+    /// Make a garbage object that will later be dropped and freed.
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `count`
+    /// elements of type `T`.
+    ///
+    /// Note: The object must be `Send + 'self`.
```
Vtec234: Same as above regarding `'self`.

src/garbage.rs
```diff
@@ -0,0 +1,140 @@
...
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `size`
+    /// elements of type `T`.
+    ///
+    /// Note: The object must be `Send + 'self`.
+    pub fn new_destroy<T>(object: *mut T, size: usize, destroy: unsafe fn(*mut T, usize)) -> Self {
+        Garbage::Destroy {
+            object: object as *mut u8,
+            size: size,
+            destroy: unsafe { mem::transmute(destroy) },
```
Vtec234: Pointers to different data types are _not_ guaranteed to be of the same size (even though they usually are). Hence, transmuting `fn(*mut T, usize)` to `fn(*mut u8, usize)` might produce wrong results.

src/garbage.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub fn new_free<T>(object: *mut T, size: usize) -> Self {
+        Garbage::Free {
+            object: object as *mut u8,
+            size: mem::size_of::<T>() * size,
+        }
+    }
+
+    /// Make a garbage object that will later be dropped and freed.
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `count`
```
Vtec234: `count` -> `size`

src/global.rs
```diff
@@ -19,15 +27,41 @@ use sync::list::List;
...
     pub static ref REGISTRIES: List<LocalEpoch> = List::new();
+    /// GARBAGES is a reference to the global queue of garbages.
+    pub static ref GARBAGES: Queue<(usize, Bag)> = Queue::new();
     /// EPOCH is a reference to the global epoch.
     pub static ref EPOCH: Epoch = Epoch::new();
 }
 
 
+#[inline]
+pub fn push_bag<'scope>(bag: &mut Bag, scope: &'scope Scope) {
```
Vtec234: Maybe add a comment stating what this does? (pushes bag onto global queue and replaces the mutref'd argument with a new bag)

src/global.rs
```diff
@@ -19,15 +27,41 @@ use sync::list::List;
...
+    let condition = |bag: &(usize, Bag)| {
+        // A pinned thread can witness at most one epoch advancement. Therefore, any bag that is
+        // within one epoch of the current one cannot be destroyed yet.
+        let diff = epoch.wrapping_sub(bag.0);
+        cmp::min(diff, 0usize.wrapping_sub(diff)) > 2
+    };
+
+    let garbages = &GARBAGES;
+    for _ in 0..COLLECT_STEPS {
+        match garbages.try_pop_if(&condition, scope) {
```
Vtec234, line 61: I really like this way of filtering over the queue :)

src/util/scoped.rs
```diff
@@ -0,0 +1,361 @@
...
+/// Creating and using a scope:
+///
+/// ```
+/// crossbeam_epoch::util::scoped::scope(|scope| {
+///     scope.defer(|| println!("Exiting scope"));
+///     scope.spawn(|| println!("Running child thread in scope"))
+/// });
+/// // Prints messages in the reverse order written
+/// ```
+pub fn scope<'a, F, R>(f: F) -> R
```
Vtec234: I hope that adding this here can be a temporary solution until scoped threads are made into a proper crate or readded to the standard library.

Vtec234 (changes requested):
Thanks a lot for working on this :) Mostly nitpicks from me.

src/global.rs
jeehoonkang: I just recalled that @stjepang commented on it: https://github.com/crossbeam-rs/crossbeam-epoch/pull/3#issuecomment-318157956

In short, it's wrong..  I will read how `coco` dealt with it, and think how to resolve this issue here.

jeehoonkang:
I addressed @Vtec234 's comments.  Thank you!

There is an unresolved issue on `ms_queue::Queue::try_pop_if()`: https://github.com/crossbeam-rs/crossbeam-epoch/pull/5#discussion_r134118438  I will work on it.

Also I am worrisome about the benchmark result: it is consistently slower than `coco`.  Yet I prefer to we can deal with it in a later PR.  What do you think?

src/sync/ms_queue.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// Attempt to dequeue from the front, if the item satisfies the given condition.
+    ///
+    /// Returns `None` if the queue is observed to be empty, or the head does not satisfy the given
+    /// condition.
+    pub fn try_pop_if<F>(&self, condition: F, scope: &Scope) -> Option<T>
+    where
+        F: Fn(&T) -> bool,
+    {
+        loop {
+            if let Ok(head) = self.pop_internal(scope) {
```
Vtec234: Hm, I don't see how this implementation can get into an unsafe situation given that it evaluates the `condition` _after_ the element is popped and is thus almost equivalent to `let a = try_pop(); if cond(a) { return a; }`. Granted, this behaviour does not actually meet the specification and is more like `try_pop_and_then_return_if`. Especially the leaking of elements on unmet `condition` is problematic, since we forget tons of garbage `Bag`s.

ghost:
I'd say don't worry about performance for now, but if you're looking for clues: it's possible that the queue is slow. IIRC, `MsQueue` is not particularly efficient, and in Coco I reimplemented the queue from scratch with slightly different memory orderings that made it a bit faster.

jeehoonkang:
@Vtec234 @stjepang I added a commit that solves ms_queue's `try_pop_if` problem.  It just follows the design of `coco`.

src/mutator.rs
```diff
@@ -254,7 +254,7 @@ impl Scope {
     }
 
     /// Deferred destruction and deallocation of heap-allocated object `ptr`.
-    pub unsafe fn defer_drop<T: Send + 'static>(&self, ptr: Ptr<T>) {
+    pub unsafe fn defer_drop<'s, T: Send + 's>(&'s self, ptr: Ptr<'s, T>) {
```
Vtec234: If we want to allow arbitrary lifetimes, we should warn the programmer about the possible pitfalls of not using `Send + 'static` data in the function comment, as described [here](https://github.com/crossbeam-rs/crossbeam/pull/57#issuecomment-197055562) and generally in that thread. I'm still in favour of only allowing `Send + 'static`, but I understand that it might be too restrictive.

src/garbage.rs
jeehoonkang: It turns out that `'static` is too restrictive.  For example, in `rsdb`, a method `defer_drop` a `Ptr` that does not have `'static` lifetime.  See https://github.com/spacejam/rsdb/issues/89 for more details.

src/mutator.rs
jeehoonkang: I agree with you..  It seems a tricky problem...

src/garbage.rs
```diff
@@ -0,0 +1,143 @@
...
+    ///
+    /// The specified object is an array allocated at address `object` and consists of `size`
+    /// elements of type `T`.
+    ///
+    /// Note: The object must be `Send + 'static`.
+    pub fn new_destroy<T>(object: *mut T, size: usize, destroy: unsafe fn(*mut T, usize)) -> Self {
+        Garbage::Destroy {
+            object: object as *mut u8,
+            size: size,
+            // FIXME(jeehoonkang): here we unsafely assume that `fn(*mut T, usize)` and `fn(*mut u8,
```
ghost, line 61: I don't really see how you can fix this. :) FWIW, Rayon uses the same assumption all over the place. I think it's a pretty safe one.

src/garbage.rs
```diff
@@ -0,0 +1,143 @@
...
+use arrayvec::ArrayVec;
+
+/// Maximum number of objects a bag can contain.
+#[cfg(not(feature = "strict_gc"))]
+const MAX_OBJECTS: usize = 64;
+#[cfg(feature = "strict_gc")]
+const MAX_OBJECTS: usize = 4;
+
+
+pub enum Garbage {
```
ghost, line 37: "Garbage" is typically in plural and refers to a bunch of deleted/dead/unlinked objects. I'd prefer to use a better synonym in singular form. That said, I don't have any ideas for a better word to use. :)

Suggestions?

src/garbage.rs
```diff
@@ -0,0 +1,143 @@
...
+    Destroy {
+        object: *mut u8,
+        size: usize,
+        destroy: unsafe fn(*mut u8, usize),
+    },
+    Free { object: *mut u8, size: usize },
+    Fn { f: Option<SendBoxFnOnce<(), ()>> },
```
ghost, line 44: Why is this an `Option` and not just `SendBoxFnOnce`?

src/global.rs
```diff
@@ -49,6 +84,31 @@ pub fn is_pinned() -> bool {
...
+/// Just like with the safe epoch::pin function, unprotected use of atomics is enclosed within a
+/// scope so that pointers created within it don't leak out or get mixed with pointers from other
+/// scopes.
+pub unsafe fn unprotected<F, R>(f: F) -> R
+where
+    F: FnOnce(&Scope) -> R,
+{
+    let mut bag = Bag::new();
+    let result = unprotected_with_bag(&mut bag, f);
+    drop(bag);
```
ghost: This drop is unnecessary.

src/util/scoped.rs
ghost: Yeah, we should really move them into a separate crate, perhaps `crossbeam-scope` (or `crossbeam-scoped`, not sure).
Shall we create this repository and start moving scoped threads there?

src/mutator.rs
ghost: Yeah, I'd rather use `Send + 'static` here. Is there a particular example of this type bound being too cumbersome to deal with?

ghost (commented):
Sorry for taking so long to review this!

src/garbage.rs
Vtec234: Admittedly, this is only the case for more exotic platforms like DSPs, but we can still leave a short comment stating the assumption.

src/util/scoped.rs
jeehoonkang: I am thinking of the crate's name.  `scoped.rs` depends on `atomic_option.rs`, so it will provide more functionalities than just scoped threads.  Maybe we can name it `crossbeam-util` or `crossbeam-base`.  What do you think?

src/mutator.rs
jeehoonkang: I saw `rsdb` needs more relaxed lifetime requirement here: https://github.com/spacejam/rsdb/blob/master/src/stack.rs#L158  I think it is exactly the case that is discussed in the issue @Vtec234 referenced.

I am up for using `send + 'static` now.  I think we need an RFC for dealing with this issue.  I will update this PR.

src/garbage.rs
jeehoonkang: What about `Trash` or `Debris`?

src/garbage.rs
jeehoonkang: In order to take `SendBoxFnOnce` object from `Garbage` in `Garbage::drop`.  `SendBoxFnOnce` doesn't provide `into` method for that purposes.

Since it is implemented as`Box<...>`, `Option<SendBoxFnOnce>` is effectively `Option<Box<...>>`, which I hope Rust will optimize out the space overhead of `Option` in codegen.

jeehoonkang:
I addressed @stjepang's comments.  Thank you!

src/garbage.rs
Vtec234: It seems that such a word literally doesn't exist: [link](https://english.stackexchange.com/questions/240876/single-countable-word-for-trash-garbage). The closest name is `PieceOfGarbage`.

src/epoch.rs
```diff
@@ -1,19 +1,19 @@
...
+//! mutator will hold a reference to it. That is the crux of safe memory reclamation.
 
 use std::ops::Deref;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::{Relaxed, Acquire, Release, SeqCst};
 
 use mutator::LocalEpoch;
 use mutator::Scope;
-use sync::list::{List, IterResult};
+use sync::michael_list::{List, IterResult};
```
ghost: Honestly, I dislike identifiers like `michael_list`, `TreiberStack`, `MsQueue` and `chase_lev`:

1. Some of these data structures (particularly list and stack) are so simple that associating them with just one of the papers discussing them is sort of odd.
2. Our implementations are not really identical to the ones presented in papers. 
3. `list`, `Stack`, `Queue`, `deque` are easier names to type and remember.
4. Choosing a less specific name doesn't lock you in to a particular implementation. E.g. if you implement a `MsQueue`, but a month later find out how to change the algorithm to be faster, it will probably stat diverging from the original paper by Michael & Scott, but the name will stay `MsQueue`.

src/util/scoped.rs
ghost: Well, it doesn't need to provide more functionality than just scoped threads: `crossbeam-scoped` can simply depend on `crossbeam-option`.

That said, it might make sense to have something like `crossbeam-utils` instead of a bunch of tiny crates.

src/garbage.rs
jeehoonkang: It seems so..  What about just `Piece`?  I think it is not unclear in this context.

src/util/scoped.rs
jeehoonkang: I prefer `crossbeam-utils`: I'm worried of too many crates.

I made `crossbeam-utils` in my account (https://github.com/jeehoonkang/crossbeam-utils), and made this PR to use it.  If you all agree with this, I will move it to `crossbeam-rs`, and register it to crates.io.

src/epoch.rs
jeehoonkang: I renamed them to `list` and `queue`.

src/garbage.rs
arthurprs: I'm fine with `Garbage`, `Object`, `Item`, .. it's not part of the surface area anyway.

src/global.rs
arthurprs: This is still pending, right?

arthurprs (commented):
It's looking good. Let's get it merged before doing improvements.

src/garbage.rs
ghost: I don't have a good alternative either. :) Let's leave it like this, at least for now.

src/global.rs
Vtec234: @arthurps: https://github.com/crossbeam-rs/crossbeam-epoch/pull/5/commits/b331551030800bfd7618db3e4b7f7055126c43ea fixed the problem

Vtec234 (approved):

src/global.rs
arthurprs: Awesome, thanks.

arthurprs (approved):
:tada:

src/epoch.rs
jeehoonkang: Honestly, I don't have any idea.  I just cut-and-pasted https://github.com/stjepang/coco/blob/master/src/epoch/thread.rs#L187 .  @stjepang any thoughts?

ghost (approved):

jeehoonkang:
I am merging it on the behalf of 3 approvals.  Thank you all for reviewing!


There are several pending issues, however:

- `lazy_static` is blocking.
- `fn defer_drop()` requires the types to be `'static`.  It may be too restrictive.
- Why `#[cold]` in `fn try_advance()`?
- Can we think of a better name for `Garbage`?

I am filing issues for each of them so that we can address them later.

jeehoonkang:
I accidentally added `garbage.rs.bk` in this branch.  I just pushed a commit that removes it.  I am sorry.

PR #7, opened 2017-08-16
Allow and truncate big tags in atomic's fetch_*

### LINKED ISSUE #6: Relaxing the condition on pointer tags
I am porting @stjepang 's `channel` (https://github.com/stjepang/channel) from `coco` to `crossbeam-epoch`, and found out that `channel` uses pointer tags `>= 8` on x86-64: https://github.com/stjepang/channel/blob/master/src/flavors/list.rs#L102

It turns out that `coco` allows tags `>= 8`, while `crossbeam-epoch` doesn't: https://github.com/stjepang/coco/blob/master/src/epoch/atomic.rs#L429 https://github.com/crossbeam-rs/crossbeam-epoch/blob/master/src/atomic.rs#L518

I am convinced that this relaxation of `coco` is actually more ergonomic, as evidenced by the channel implementation.  Though it may be a little bit more unsafe (in terms of surprising the users).  In this tradeoff, I would like to prefer the usability to the safety.  What do you think?

Also, if we will decide to be strict on pointer tags in this matter, I think it would be great to expose `atomic.rs`'s `low_bits()` in the public interface, for supporting the use case in `channel`.

jeehoonkang:
Currently, `fetch_add`, `fetch_or`, `fetch_xor` for atomics require that the pointer tags are fit in the unused bits in a pointer.  However, this may be too restrictive for users, as evidenced by the use of big tags in @stjepang's `channel` (https://github.com/stjepang/channel/blob/master/src/flavors/list.rs#L102).

This PR allows big tags, but truncate them before actually using them, as `coco` does (https://github.com/stjepang/coco/blob/master/src/epoch/atomic.rs#L429).  In order for the user to be able to check if their tags are fit in the unused bits, we export the `tag_bits()` function that lets you know which bits are usable in tags.

This PR is currently not `cargo test`able.  But I manually tested it on top of #5, and it worked fine.  Also, `channel`'s `cargo test` also worked fine: https://github.com/jeehoonkang/channel/tree/crossbeam-epoch

Closes #6

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -496,7 +496,8 @@ impl<T> Atomic<T> {
     /// Bitwise "and" with the current tag.
     ///
     /// Performs a bitwise "and" operation on the current tag and the argument `val`, and sets the
-    /// new tag to the result. Returns the previous pointer.
+    /// new tag to the result. Returns the previous pointer.  Before the operation, `val` is
+    /// truncated to be fit in the bitmask returned by [`tag_bits()`](fn.tag_bits.html)).
     ///
     /// This method takes an [`Ordering`] argument which describes the memory ordering of this
     /// operation.
@@ -522,7 +523,8 @@ impl<T> Atomic<T> {
     /// Bitwise "or" with the current tag.
     ///
     /// Performs a bitwise "or" operation on the current tag and the argument `val`, and sets the
-    /// new tag to the result. Returns the previous pointer.
+    /// new tag to the result. Returns the previous pointer.  Before the operation, `val` is
+    /// truncated to be fit in the bitmask returned by [`tag_bits()`](fn.tag_bits.html)).
     ///
     /// This method takes an [`Ordering`] argument which describes the memory ordering of this
     /// operation.
@@ -548,7 +550,8 @@ impl<T> Atomic<T> {
     /// Bitwise "xor" with the current tag.
     ///
     /// Performs a bitwise "xor" operation on the current tag and the argument `val`, and sets the
-    /// new tag to the result. Returns the previous pointer.
+    /// new tag to the result. Returns the previous pointer.  Before the operation, `val` is
+    /// truncated to be fit in the bitmask returned by [`tag_bits()`](fn.tag_bits.html)).
     ///
     /// This method takes an [`Ordering`] argument which describes the memory ordering of this
     /// operation.
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -68,13 +68,13 @@ impl CompareAndSetOrdering for (Ordering, Ordering) {
 /// Panics if the pointer is not properly unaligned.
 #[inline]
 fn ensure_aligned<T>(raw: *const T) {
-    assert_eq!(raw as usize & tag_bits::<T>(), 0, "unaligned pointer");
+    assert_eq!(raw as usize & low_bits::<T>(), 0, "unaligned pointer");
 }
 
 /// Panics if the tag doesn't fit into the unused bits of an aligned pointer to `T`.
 #[inline]
 fn validate_tag<T>(tag: usize) {
-    let mask = tag_bits::<T>();
+    let mask = low_bits::<T>();
     assert!(
         tag <= mask,
         "tag too large to fit into the unused bits: {} > {}",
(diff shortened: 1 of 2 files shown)


ghost (approved):
Looks good to me!

Yeah, I think the right thing to do here is simply skip validation of tags for bitwise operations.

src/atomic.rs
```diff
@@ -516,8 +516,7 @@ impl<T> Atomic<T> {
     /// });
     /// ```
     pub fn fetch_and<'scope>(&self, val: usize, ord: Ordering, _: &'scope Scope) -> Ptr<'scope, T> {
-        validate_tag::<T>(val);
-        Ptr::from_data(self.data.fetch_and(val, ord))
+        Ptr::from_data(self.data.fetch_and(val | !tag_bits::<T>(), ord))
     }
 
     /// Bitwise "or" with the current tag.
```
Vtec234: Please mention in the documentation of all three functions that the tags are truncated to a valid size.

Vtec234 (changes requested):
LGTM, just please update the documentation to reflect this so that the behaviour isn't entirely unexpected.

jeehoonkang:
@Vtec234 Thank you for the comments.  I just updated the PR.

Vtec234 (approved):

src/atomic.rs
```diff
@@ -496,7 +496,8 @@ impl<T> Atomic<T> {
     /// Bitwise "and" with the current tag.
     ///
     /// Performs a bitwise "and" operation on the current tag and the argument `val`, and sets the
-    /// new tag to the result. Returns the previous pointer.
+    /// new tag to the result. Returns the previous pointer.  Before the operation, `val` is
+    /// truncated to be fit in the bitmask returned by [`tag_bits()`](fn.tag_bits.html)).
```
ghost: I'd rather not mention the `tag_bits()` method since it is not public.

My suggestion is to leave this comment as-is, and simply add the following before `# Examples` in the documentation for `Owned::with_tag` and `Ptr::with_tag`:

```rust
/// # Panics
///
/// Panics if the tag doesn't fit into the least significant unused bits of an aligned pointer to `T`.
```

@Vtec234  said:

> Please mention in the documentation of all three functions that the tags are truncated to a valid size.

But the documentation here already states: *Performs a bitwise "and" operation on the current tag and the argument `val`*, so I think truncation is already implied.

ghost: Oh wait, `tag_bits` *is* public now? Shouldn't it be just a purely internal helper function?

Vtec234: I guess it implies truncation for `and`, but it doesn't for `or`, since that wouldn't truncate it. In any case, it's not immediately obvious, so it's better to explicitly state that this is what happens. `tag_bits` being public allows the user to check that their tag fits if they want to make sure it does.

ghost: I see. In that case, my only complaints would be:

1. Parentheses don't match up: `(fn.tag_bits.html))` :) 
2. "to be fit in" -> "to fit into"

@jeehoonkang  You can merge the PR after that.

Vtec234: Hm, but actually since we say that it's the alignment, the user can just check for `tag < align_of::<T>()` rather than `tag <= tag_bits()`, so do we need to expose it?

ghost: Yeah... I don't think it's necessary.

jeehoonkang: I left the comments as is, and add a comment in the type definition: https://github.com/crossbeam-rs/crossbeam-epoch/pull/7/files#diff-422fa30cb418a44869868fb187a6ded0R102

jeehoonkang:
Also, I think it is consistent to allow big tags in initializer: https://github.com/crossbeam-rs/crossbeam-epoch/pull/7/files#diff-422fa30cb418a44869868fb187a6ded0R96

ghost:
It seems that `validate_tag` is now an unused function - should we remove it?

Also, let's document in `Ptr::with_tag` and `Owned::tag` that the tag passed as the argument will be truncated to fit into the unused bits.

jeehoonkang:
@stjepang Indeed, good catch! I just removed `validate_tag()`.  I add a comment in `Ptr::with_tag()`.  But for `Owned::tag()`, `tag` is not an argument, but rather a return value. So I didn't write a comment on it.  Thanks!

ghost:
Sorry, I meant `Owned::with_tag` :)

jeehoonkang:
Yeah, I updated it :)

PR #8, opened 2017-08-24
Impl Drop for Owned

ghost:
`Owned<T>` is just a `Box<T>` that also allows pointer tagging.

It is a smart pointer so it needs to automatically destroy the pointer it owns.

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -716,6 +716,15 @@ impl<T> Owned<T> {
     }
 }
 
+impl<T> Drop for Owned<T> {
+    fn drop(&mut self) {
+        let raw = (self.data & !low_bits::<T>()) as *mut T;
+        unsafe {
+            drop(Box::from_raw(raw));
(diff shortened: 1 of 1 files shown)


Vtec234:
Adding this would require making `from_raw` as well as `from_data` unsafe, otherwise `Owned::from_raw(0 as *mut u32);` is safe code that segfaults when it drops.

ghost:
@Vtec234 Oh, I didn't notice that, thanks!

It's funny that the docs for `Owned::from_raw` already had the following sentence:

> This function is unsafe because improper use may lead to memory problems. For example, a double-free may occur if the function is called twice on the same raw pointer.

I've added a new commit that makes this function unsafe.

Vtec234 (approved):
I think `from_data` should also be unsafe, even though it's private. Otherwise, LGTM.

jeehoonkang (approved):

Issue #10, opened 2017-08-28
`lazy_static` is blocking


jeehoonkang:
As discussed in https://github.com/crossbeam-rs/crossbeam/pull/122#issuecomment-288436809, `lazy_static` is blocking.  It is used for the global registries, the garbage queue, and the global epoch.  It would be great if accessing them is non-blocking.

arthurprs:
It's only blocking in the slow path due to Once, it only does a seqcst load on the fast-path  (https://github.com/rust-lang/rust/blob/c7104be1a37daa2f7ec114913247dc93524fd48e/src/libstd/sync/once.rs#L220)

arthurprs:
Should we close this? (based on the above comment)

jeehoonkang:
I feel we cannot easily change the current state of affair.  I'm closing this now, but please feel free to reopen it.

PR #13, opened 2017-09-06
Update the implementation of unprotected scope

jeehoonkang:
This is a subset of RFC #10 (https://github.com/crossbeam-rs/rfcs/pull/10).  It doesn't change any interface, so I would like to post this PR without going for RFC.

In short, it implements @stjepang 's suggestion (https://github.com/crossbeam-rs/rfcs/pull/10#issuecomment-327125792): `unprotected` scope has a null bag, and garbages created within this scope are immediately disposed of.

### DIFF
--- a/src/mutator.rs
+++ b/src/mutator.rs
@@ -16,6 +16,7 @@
 //! are necessary for performing atomic operations, and for freeing/dropping locations.
 
 use std::cell::{Cell, UnsafeCell};
+use std::ptr;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::{Relaxed, Release, SeqCst};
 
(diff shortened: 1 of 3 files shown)


src/mutator.rs
```diff
@@ -226,17 +238,22 @@ impl LocalEpoch {
 }
 
 impl Scope {
-    unsafe fn get_bag(&self) -> &mut Bag {
-        &mut *self.bag
+    #[inline]
+    unsafe fn get_bag(&self) -> Option<&mut Bag> {
```
ghost: This can be just:

```rust
#[inline]
unsafe fn get_bag(&self) -> Option<&mut Bag> {
    self.bag.as_mut()
}
```

ghost (approved):
Nice! I have just one minor nit.

src/mutator.rs
ghost: Or, alternatively, delete this method and instead of calling `self.get_bag()` just write `self.bag.as_mut()` :)

arthurprs (approved):
Looks good.

jeehoonkang:
@stjepang thanks!  I removed `get_bag()`, as you advised.

PR #15, opened 2017-09-19
Replace SendBoxFnOnce with Deferred

ghost:
This is an implementation of [this RFC](https://github.com/crossbeam-rs/rfcs/pull/9).
I've intentionally left out the part about removing `defer_free` - this will be done in a separate PR.

### DIFF
--- a/src/deferred.rs
+++ b/src/deferred.rs
@@ -1,58 +1,20 @@
 use std::mem;
 use std::ptr;
 
-/// Provides methods to dispatch a call to a `FnOnce()` from a trait object.
-pub trait Callback {
-    /// Calls the function from a trait object on the stack.
-    ///
-    /// This will copy `self`, call the function, and finally drop the copy.
-    /// This method may be called only once, and `self` must not be dropped after that (tip: pass
-    /// it to `std::mem::forget`).
-    unsafe fn copy_and_call(&self);
-
-    /// Calls the function from a trait object on the heap.
-    fn call_box(self: Box<Self>);
-}
-
-impl<F: FnOnce()> Callback for F {
-    #[inline]
-    unsafe fn copy_and_call(&self) {
-        let f: Self = ptr::read(self);
-        f();
-    }
-
-    #[inline]
-    fn call_box(self: Box<Self>) {
-        let f: Self = *self;
-        f();
-    }
-}
-
-/// The representation of a trait object like `&SomeTrait`.
+/// Number of words a piece of `Data` can hold.
 ///
-/// This struct has the same layout as types like `&SomeTrait` and `Box<AnotherTrait>`.
-///
-/// It is actually already provided as `std::raw::TraitObject` gated under the nightly `raw`
-/// feature. But we don't use nightly Rust, so the struct was simply copied over into Crossbeam.
-///
-/// If the layout of this struct changes in the future, Crossbeam will break, but that is a fairly
-/// unlikely scenario.
-// FIXME(stjepang): When feature `raw` gets stabilized, use `std::raw::TraitObject` instead.
-#[repr(C)]
-#[derive(Copy, Clone)]
-struct TraitObject {
-    data: *mut (),
-    vtable: *mut (),
-}
+/// Three words should be enough for the majority of cases. For example, you can fit inside it the
+/// function pointer together with a fat pointer representing an object that needs to be destroyed.
+const DATA_WORDS: usize = 3;
 
 /// Some space to keep a `FnOnce()` object on the stack.
-type Data = [usize; 3];
+type Data = [usize; DATA_WORDS];
 
 /// A `FnOnce()` that is stored inline if small, or otherwise boxed on the heap.
 ///
 /// This is a handy way of keeping an unsized `FnOnce()` within a sized structure.
 pub struct Deferred {
-    vtable: *mut (),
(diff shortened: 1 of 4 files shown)


src/deferred.rs
```diff
@@ -0,0 +1,166 @@
...
+// FIXME(stjepang): When feature `raw` gets stabilized, use `std::raw::TraitObject` instead.
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct TraitObject {
+    data: *mut (),
+    vtable: *mut (),
+}
+
+/// Some space to keep a `FnOnce()` object on the stack.
+type Data = [usize; 3];
```
jeehoonkang: `3` here is a magic number.  Maybe have a `const` for it?  Also, I would like to ask how `3` is chosen here.

src/deferred.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        unsafe {
+            if size <= mem::size_of::<Data>() && align <= mem::align_of::<Data>() {
+                let vtable = {
+                    let callback: &Callback = &f;
+                    let obj: TraitObject = mem::transmute(callback);
+                    obj.vtable
+                };
+
+                let mut data = Data::default();
```
jeehoonkang: maybe we can use `mem::uninitialized()`?

src/deferred.rs
```diff
@@ -0,0 +1,166 @@
...
+                }
+            }
+        }
+    }
+
+    /// Calls the function or panics if it was already called.
+    #[inline]
+    pub fn call(&mut self) {
+        let vtable = mem::replace(&mut self.vtable, ptr::null_mut());
+        assert!(!vtable.is_null(), "cannot call `FnOnce` more than once");
```
jeehoonkang: I think this syntax requires a recent version of Rust, right?  Maybe we have to document the minimum required version of Rust..

src/deferred.rs
```diff
@@ -0,0 +1,166 @@
+use std::mem;
+use std::ptr;
+
+/// Provides methods to dispatch a call to a `FnOnce()` from a trait object.
+pub trait Callback {
+    /// Calls the function from a trait object on the stack.
```
jeehoonkang: why do we need copy here?

src/deferred.rs
```diff
@@ -0,0 +1,166 @@
+use std::mem;
+use std::ptr;
+
+/// Provides methods to dispatch a call to a `FnOnce()` from a trait object.
```
jeehoonkang: Why do we need `Callback` trait`?  Can we inline the implementation for `FnOnce` in the definition of `Deferred`?

ghost: Indeed - it's not needed. I've simplified the implementation by removing the trait.

src/deferred.rs
ghost: I've removed this. :)

src/deferred.rs
ghost: I've added a constant with some explanation.

src/deferred.rs
ghost: Yeah, that may give us some perf improvements.

src/deferred.rs
ghost: Hm, I can recall *some* recent tweaks to commonly used macros, but I'm pretty sure `assert!` has supported this syntax since forever (Rust 1.0).

But yeah, documenting the minimum required version of Rust would be nice. Do you know of any popular crates that do that (besides Rayon - I know of that one)? I'm wondering how others handle this issue.

src/deferred.rs
```diff
@@ -0,0 +1,140 @@
...
+use std::ptr;
+
+/// Number of words a piece of `Data` can hold.
+///
+/// Three words should be enough for the majority of cases. For example, you can fit inside it the
+/// function pointer together with a fat pointer representing an object that needs to be destroyed.
+const DATA_WORDS: usize = 3;
+
+/// Some space to keep a `FnOnce()` object on the stack.
+type Data = [usize; DATA_WORDS];
```
jeehoonkang, line 11: What about inlining this definition?

src/deferred.rs
```diff
@@ -0,0 +1,140 @@
...
+
+/// A `FnOnce()` that is stored inline if small, or otherwise boxed on the heap.
+///
+/// This is a handy way of keeping an unsized `FnOnce()` within a sized structure.
+pub struct Deferred {
```
jeehoonkang, line 16: There are two variants of `Deferred`: one using `data` as the function, and the other using `data` as a pointer to the function.  What about giving two variants to `enum Deferred`?

ghost: That would increase the size of `Deferred` by one additional word (to encode the discriminant) and hurt performance pretty badly. Remember - it actually used to be an enum before, but fared poorly in @Vtec234's benchmarks. To fix performance, I had to get rid of the discrimant.

src/deferred.rs
ghost: You mean, why not remove this type? Because then things like
```rust
if size <= mem::size_of::<Data>() && align <= mem::align_of::<Data>() {
```

would become
```rust
if size <= mem::size_of::<[usize; DATA_WORDS]>() && align <= mem::align_of::<[usize; DATA_WORDS]>() {
```

jeehoonkang (approved):

PR #16, opened 2017-09-21
Add Ptr::into_owned method

ghost:
This method drops and deallocates the object pointed to.

cc crossbeam-rs/rfcs#11

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -958,6 +958,38 @@ impl<'scope, T> Ptr<'scope, T> {
         self.as_raw().as_ref()
     }
 
+    /// Drops and deallocates the pointee.
+    ///
+    /// Calling this method on `p` is equivalent to:
+    ///
+    /// ```no_compile
+    /// drop(Box::from_raw(p.as_raw() as *mut T));
+    /// ```
+    ///
+    /// # Safety
+    ///
+    /// This method may be called only if the pointer is valid and the object pointed to won't be
+    /// accessed again in the future. Of course, calling this method on the same object twice is
+    /// illegal.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_epoch::{self as epoch, Atomic};
+    /// use std::sync::atomic::Ordering::SeqCst;
+    ///
+    /// let a = Atomic::new(1234);
+    /// unsafe {
+    ///     epoch::unprotected(|scope| {
+    ///         let p = a.load(SeqCst, scope);
+    ///         p.destroy();
+    ///     });
+    /// }
+    /// ```
+    pub unsafe fn destroy(self) {
+        drop(Box::from_raw(self.as_raw() as *mut T));
+    }
+
     /// Returns the tag stored within the pointer.
     ///
     /// # Examples
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -784,8 +784,10 @@ impl<T> AsMut<T> for Owned<T> {
 #[derive(Debug)]
 pub struct Ptr<'scope, T: 'scope> {
     data: usize,
-    _marker: PhantomData<&'scope T>,
+    _marker: PhantomData<(&'scope (), *const T)>,
 }
+
+unsafe impl<'scope, T: Send> Send for Ptr<'scope, T> {}
 
 impl<'scope, T> Clone for Ptr<'scope, T> {
     fn clone(&self) -> Self {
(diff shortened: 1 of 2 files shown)


jeehoonkang:
How about updating existing codes to use `Ptr::destroy()` rather than dropping-after-boxing?  For example: https://github.com/crossbeam-rs/crossbeam-epoch/blob/master/src/sync/list.rs#L123

src/atomic.rs
```diff
@@ -958,6 +958,38 @@ impl<'scope, T> Ptr<'scope, T> {
         self.as_raw().as_ref()
     }
 
+    /// Drops and deallocates the pointee.
+    ///
+    /// Calling this method on `p` is equivalent to:
+    ///
+    /// ```no_compile
+    /// drop(Box::from_raw(p.as_raw() as *mut T));
```
ghost: Someone might be asking themselves:
*Why do we allocate objects using `Owned`, but then drop them using `Box::from_raw`*?

I wonder if we should write this instead to avoid confusion:

```rust
/// drop(Owned::from_raw(p.as_raw() as *mut T));
```

ghost: And that gets me thinking: We have `Owned::into_ptr` and `Ptr::destroy`, which are symmetrical functions going in opposite directions.

Should we perhaps name them `Owned::into_ptr` and `Ptr::into_owned`, with the following definition?
```rust
impl<'scope, T> Ptr<'scope, T> {
    unsafe fn into_owned(self) -> Owned<T> { ... }
}
```

Then you can defer destruction of an object pointed to by a `Ptr` using:
```rust
scope.defer(move || drop(p.into_owned()));
```

A bit more verbose, but perhaps not too bad?

cc @arthurprs - you might have an opinion on all this. :)

Vtec234: I like this idea, because when you have a `Ptr` you can use `into_owned`, and when you have a pointer you can use `drop(Owned::from_ptr(..))`. Then we could indeed remove `defer_drop`. Moreover, it clearly states that we are retaking ownership of the pointer and the right to destroy it in the future. I would prefer it to adding `Ptr::destroy`.

arthurprs: `Ptr::into_owned` sounds way better imho. This is probably enough though `scope.defer(move || p.into_owned());`, right? Looks fairly compact.

ghost: @arthurprs 
```rust
scope.defer(move || drop(p.into_owned())); // OK
scope.defer(move || { p.into_owned(); }); // OK
scope.defer(|| drop(p.into_owned())); // Compiles, but I strongly encourage *always* using move closures.
scope.defer(move || p.into_owned()); // Error: expected (), found struct `atomic::Owned`
```

src/atomic.rs
```diff
@@ -958,6 +958,38 @@ impl<'scope, T> Ptr<'scope, T> {
...
+    /// let a = Atomic::new(1234);
+    /// unsafe {
+    ///     epoch::unprotected(|scope| {
+    ///         let p = a.load(SeqCst, scope);
+    ///         p.destroy();
+    ///     });
+    /// }
+    /// ```
+    pub unsafe fn destroy(self) {
+        drop(Box::from_raw(self.as_raw() as *mut T));
```
jeehoonkang: Then it might be better to replace this line with:

```rust
drop(Owned::from_raw(self.as_raw() as *mut T));
```

jeehoonkang (approved):
I like both of @stjepang's idea (one on comment, the other on `into_owned()`).  I'm approving this PR :smile:

Vtec234 (approved):

src/atomic.rs
```diff
@@ -784,9 +784,11 @@ impl<T> AsMut<T> for Owned<T> {
 #[derive(Debug)]
 pub struct Ptr<'scope, T: 'scope> {
     data: usize,
-    _marker: PhantomData<&'scope T>,
+    _marker: PhantomData<(&'scope (), *const T)>,
```
jeehoonkang, line 787: What's the difference here?

src/atomic.rs
```diff
@@ -784,9 +784,11 @@ impl<T> AsMut<T> for Owned<T> {
...
 }
 
+unsafe impl<'scope, T: Send> Send for Ptr<'scope, T> {}
```
jeehoonkang, line 790: Do we need this?

src/atomic.rs
arthurprs: The raw ptr will make it non Send by default.

arthurprs: 3 lines bellows it's made Send again but only for T: Send.

src/atomic.rs
Vtec234: Yes, because `defer` requires `Send`, so for `Ptr` to be usable with `defer` we do need it. That said, raw pointer aren't `Send`, so maybe this is a bit wierd, but it's meant to be used in concurrent contexts.

src/atomic.rs
ghost: A `Ptr` is *not* a reference - it's more like a pointer that can be dereferenced (using `.deref()` or `.as_ref()`). Also, pointers are nullable.

Basically:

1. `_marker: PhantomData<&'scope T>` says: *I'm a shared reference to `T` tied to `'scope`.*
2. `_marker: PhantomData<(&'scope (), *const T)>` says: *I'm a shared pointer to `T` and tied to `'scope`.*

The second option describes `Ptr` better.

src/atomic.rs
ghost: Yes. Since `Ptr` now contains a `*const T`, it will not be `Send` by default and you won't be able to pass it into `scope.defer(...)`. For that reason, we must manually implement `Send`.

Issue #14, opened 2017-09-06
Announcing crossbeam-epoch 0.1.0?


jeehoonkang:
It seems crossbeam-epoch is comparable to both the original crossbeam and coco in terms of features.  For example, I could successfully port rsdb, which was based on coco, to crossbeam-epoch: https://github.com/jeehoonkang/rsdb/tree/crossbeam-epoch  Also, its performance seems not bad compared to the two precursors.

So I wonder if (1) we want to announce the initial 0.1.0 release of the core of new crossbeam, and (2) when.  Personally, I want to announce it after both https://github.com/crossbeam-rs/rfcs/pull/9 and https://github.com/crossbeam-rs/crossbeam-epoch/pull/13 are somehow resolved.

What do you think?

ghost:
Yesterday I noticed segmentation faults and would like to investigate what went wrong before releasing 0.1.0.

I think as far as features go, we're getting ready for an early release, but I'm still feeling a bit uneasy without at least some basic sanitization tests (similar to the tests I wrote for `coco`).

jeehoonkang:
I'd like to enumerate the release blockers.

- We need tests (cf. #17).

- @stjepang Can you reproduce the segfault for the current master?

- I guess @Vtec234 wants to add benchmark before we publish crossbeam-epoch 0.1.0.  But it's unclear which benchmark we should evaluate before releasing 0.1.0.

jeehoonkang:
How about writing up a post for advertising `crossbeam-epoch` for wider audience (e.g. reddit, lobste.rs) at the time of 0.1.0 release?

cf. the write-up will close https://github.com/crossbeam-rs/rfcs/issues/7

ghost:
@jeehoonkang I'll submit a PR that adds some benchmarks soon (today or tomorrow).

Regarding your comments on backward compatibility, I think it's simply inevitable that we'll break compatibility multiple times within the next several months. This is normal for such a young and complicated project. :)

I'd prefer to release a new version soon so that other subprojects (`crossbeam-deque` and `crossbeam-channel`) can start relying on it. We have to kick the ball rolling at some point.

The project seems to be in a fairly solid state now. I've just run tests on my Raspberry Pi (ARM v7) and everything seems to be working just fine. What do you think - after adding benchmarks, can we release a 0.1.0 so that `crossbeam-epoch` can be used as a dependency?

We don't have to publicise this release yet, but once things settle down, we'll definitely do it.

[...]

PR #17, opened 2017-09-22
Add tests

jeehoonkang:
Now crossbeam-epoch's test coverage is largely the same with coco.

A caveat is that we need to do `RUST_TEST_THREADS=1 cargo test` instead of `cargo test` for sequentially executing test codes.  This is because two tests (`incremental` and `buffering`) are correct under the assumption that the thread in concern is the only thread that interacts with the global garbage queue.  I think the best way to deal with this problem is using custom instances (realms) of EBR instead of the global instance in test codes.

I'd like to suggest we publish crossbeam-epoch 0.1.0 after merging this, #16, and the custom realms proposal discussed in https://github.com/crossbeam-rs/rfcs/issues/12.  I'd like to see how existing projects will adopt the new Crossbeam™.  Cf. #14.


ghost:
Regarding parallel test execution, check out [this](https://github.com/rust-lang/rust/issues/33519#issuecomment-314224553) comment. Also, it seems `RUST_TEST_THREADS=1` is [deprecated](https://stackoverflow.com/questions/44947914/how-to-limit-the-number-of-test-threads-in-cargo-toml).

I'd prefer to avoid relying on the flag for correct test execution. I think `cargo test` should *always* pass without requiring any additional flags. The recommended workaround to make tests run sequentially would be to create a mutex and lock it just before running a test.

Also, note that if you write tests in `tests/foo1.rs` and `tests/foo2.rs`, these two files will be tested one after another (but tests within each file will be run in parallel). However, all tests in `src/**` will be run concurrently.

Other than that, this looks pretty good. :) Did you port all relevant tests from coco now? Are there any we're still missing?

jeehoonkang:
- In my opinion, the test should really be run in parallel.  (After all, Crossbeam is for concurrency and parallelism!)  I think this PR can wait for the custom realm proposal to be landed, after which test code can use their own custom realm, no longer requiring sequential execution of test codes.

- I ported every test except for `garbage.rs`'s `smoke()` and `flush_pending()`.  They are testing pending bags in coco, which have no corresponding notion in the current implementation in `crossbeam-epoch`.

jeehoonkang:
Hi!  this is rebased on #21 (the `Collector` API), and thanks to it, now we can safely run tests in parallel :)  It is quite comprehensive test cases for `Collector`, which are mostly cut-and-pasted from coco.  I think it's ready to be merged.

src/sync/list.rs
```diff
@@ -93,17 +96,21 @@ impl<T> List<T> {
     }
 
     /// Inserts `data` into the head of the list.
+    #[inline]
     pub fn insert<'scope>(&'scope self, data: T, scope: &'scope Scope) -> Ptr<'scope, Node<T>> {
         Self::insert_internal(&self.head, data, scope)
     }
 
-    /// Inserts `data` after `to`.
+    /// Inserts `data` after into the head of the list.
```
Vtec234: Should this be "inserts `data` after `after` into the list."?

Vtec234 (approved):
Something funky is going on with the GitHub diff display after the rebase, but I checked it locally and it looks fine.

arthurprs (approved):
Awesome work on this refactor/rfc :tada:

src/sync/list.rs
jeehoonkang: Thanks!  I just pushed a new commit that changes the doc as you said.

src/collector.rs
```diff
@@ -0,0 +1,388 @@
...
+        static DESTROYS: AtomicUsize = ATOMIC_USIZE_INIT;
+
+        let collector = Collector::new();
+        let handle = collector.handle();
+
+        handle.pin(|scope| unsafe {
+            for _ in 0..COUNT {
+                let a = Owned::new(7i32).into_ptr(scope).as_raw() as usize;
+                scope.defer(move || {
+                    drop(Owned::from_raw(a as *mut i32));
```
ghost: I think we can use `a.into_owned()` here, and remove the `.as_raw() as usize` part above.

src/collector.rs
```diff
@@ -0,0 +1,388 @@
...
```
ghost: Same here.

ghost (approved):
Just a few nits.

jeehoonkang:
@stjepang Thanks!  I just changed the code as you said.

PR #18, opened 2017-09-23
Change the `sync::list` API

jeehoonkang:
@Vtec234, @stjepang, and I had a relevant discussion: https://github.com/crossbeam-rs/crossbeam-epoch/pull/4#discussion_r130244770 . In short, I made a PR in which `sync` was public, but finally we agree that `sync` is out of the scope of `crossbeam-epoch`.  Now I have a (partly) different opinion: `sync::list` might be a useful construction for a variety of nonblocking data structures, and it would be good to expose it.  This PR is doing that.

These days I'm implementing "A Wait-free Queue as Fast as Fetch-and-Add" ([link](http://dl.acm.org/citation.cfm?id=2851168)).  In order to achieve wait-freedom, it uses the "helping" technique: a "fast" thread that has progressed should "help" other "slow" threads to make progress.  In these constructions, it is necessary for the helping thread to examine the status of all the other threads.

For that, many wait-free data structures maintain the list of all the participating threads.  For this reason, I think `sync::list` in this PR would be fit into the scope of the `crossbeam-epoch` crate, as a library for wait-free constructions.  I adjusted `sync::list` a little bit so that wait-free data structures can easily use it.  In particular, I added `ring_iter(&self: &'scope List<T>, starter: &Node<T>, scope: &'scope Scope)` method, which returns a node iterator from the one right after `starter` to the one right before `starter`.

### DIFF
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -13,7 +13,7 @@ use std::sync::atomic::Ordering::{Relaxed, Acquire, Release, SeqCst};
 
 use mutator::LocalEpoch;
 use mutator::Scope;
-use sync::list::{List, IterResult};
+use sync::list::{List, IterError};
 use crossbeam_utils::cache_padded::CachePadded;
 
 /// The global epoch is a (cache-padded) integer.
@@ -41,16 +41,15 @@ impl Epoch {
         ::std::sync::atomic::fence(SeqCst);
 
         // Traverse the linked list of mutator registries.
-        let mut registries = registries.iter(scope);
-        loop {
-            match registries.next() {
-                IterResult::Abort => {
-                    // We leave the job to the mutator that also tries to advance to epoch and
-                    // continues to iterate the registries.
+        for registry in registries.iter(scope) {
+            match registry {
+                Err(IterError::LostRace) => {
+                    // We leave the job to the mutator that won the race, which continues to iterate
+                    // the registries and tries to advance to epoch.
                     return epoch;
                 }
-                IterResult::None => break,
-                IterResult::Some(local_epoch) => {
+                Ok(local_epoch) => {
+                    let local_epoch = local_epoch.get();
                     let (mutator_is_pinned, mutator_epoch) = local_epoch.get_state();
 
                     // If the mutator was pinned in a different epoch, we cannot advance the global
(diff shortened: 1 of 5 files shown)


ghost:
If I'm understanding correctly, `crossbeam-epoch` would have it's own list of participating threads, and the `crossbeam-list` would have another, separate list?

My first hunch is "this should really be in `crossbeam-list`". I'd be okay with having this in `crossbeam-epoch` for now and let the data structure evolve a little bit, but then later on revisit the overall API and come up with a more thoroughly thought out design. At that point I'd suggest taking into consideration at least three use cases:

1. In `crossbeam-epoch` as the list of registered threads.
2. In `crossbeam-queue` as the list for wait-free stuff.
3. In `rayon` as a possible data structure for flat combining (there was an experimental PR).

jeehoonkang:
@stjepang A problem is, `crossbeam-epoch` and `crossbeam-list` have dependency to each other: `crossbeam-list` is built on top of Crossbeam's EBR.  That's the reason I think they should be in the same crate.

We can break this dependency by introducing another crate, which we may call `crossbeam-types`, that defines the types of EBR.  `crossbeam-epoch` is an implementation of that.  Then `crossbeam-list` depends only on `crossbeam-types`, and `crossbeam-epoch` depends on both `crossbeam-list` and `crossbeam-types`.  But I think this solution is a little bit verbose.

ghost:
Yeah, that's way too much contortion just to make the dependencies work out. :)

So, my opinion is still that a well implemented `List<T>` with powerful interface (1000s of lines of code) belongs to `crossbeam-list`. This should be a one-size-fits-all linked list for everyone who needs a concurrent list.

On the other hand, `crossbeam-epoch` doesn't need anything special - just a trivial linked list in 50 lines of code or so is enough (like in `coco`). If we duplicate this very modest amount of code, it's no big deal, is it? :) After all, as we once discussed, we should also experiment with a different, more cache-friendly data structure for the thread list (or, should I say now, *handle list*).

jeehoonkang:
Yes..  I agree that handle list may have different characteristics than other lists, and for now it'd be better to implement `crossbeam-epoch`'s own list.  Thank you.

I narrowed the scope of this PR down.  It only refactors `sync::list` and gives (in my opinion) a better API for list.  The major improvement is, now it uses the standard `Iter` trait.

ghost (approved):
LGTM

Vtec234 (approved):

PR #20, opened 2017-09-25
Remove defer_free and defer_drop

ghost:
This is a follow-up to the discussion we had in #16.

### DIFF
--- a/src/mutator.rs
+++ b/src/mutator.rs
@@ -247,28 +247,8 @@ impl Scope {
         });
     }
 
-    /// Deferred deallocation of heap-allocated object `ptr`.
-    ///
-    /// This function inserts the object into a mutator-local [`Bag`]. When the bag becomes full,
-    /// the bag is flushed into the globally shared queue of bags.
-    ///
-    /// If the object is unusually large, it is wise to follow up with a call to [`flush`] so that
-    /// it doesn't get stuck waiting in the local bag for a long time.
-    ///
-    /// [`Bag`]: struct.Bag.html
-    /// [`flush`]: fn.flush.html
-    pub unsafe fn defer_free<T>(&self, ptr: Ptr<T>) {
-        self.defer_garbage(Garbage::new_free(ptr.as_raw() as *mut T, 1))
-    }
-
-    /// Deferred destruction and deallocation of heap-allocated object `ptr`.
-    // FIXME(jeehoonkang): `T: 'static` may be too restrictive.
-    pub unsafe fn defer_drop<T: Send + 'static>(&self, ptr: Ptr<T>) {
-        self.defer_garbage(Garbage::new_drop(ptr.as_raw() as *mut T, 1))
-    }
-
     /// Deferred execution of an arbitrary function `f`.
-    pub unsafe fn defer<F: FnOnce() + Send + 'static>(&self, f: F) {
+    pub unsafe fn defer<F: FnOnce() + Send>(&self, f: F) {
         self.defer_garbage(Garbage::new(f))
     }
 
--- a/src/mutator.rs
+++ b/src/mutator.rs
@@ -248,8 +248,8 @@ impl Scope {
     }
 
     /// Deferred execution of an arbitrary function `f`.
-    pub unsafe fn defer<F: FnOnce() + Send>(&self, f: F) {
-        self.defer_garbage(Garbage::new(f))
+    pub unsafe fn defer<R, F: FnOnce() -> R + Send>(&self, f: F) {
+        self.defer_garbage(Garbage::new(|| drop(f())))
     }
 
     /// Flushes all garbage in the thread-local storage into the global garbage queue, attempts to
--- a/src/mutator.rs
+++ b/src/mutator.rs
@@ -248,6 +248,15 @@ impl Scope {
     }
 
     /// Deferred execution of an arbitrary function `f`.
+    ///
+    /// This function inserts the function into a mutator-local [`Bag`]. When the bag becomes full,
+    /// the bag is flushed into the globally shared queue of bags.
+    ///
+    /// If this function is destroying a particularly large object, it is wise to follow up with a
+    /// call to [`flush`] so that it doesn't get stuck waiting in the local bag for a long time.
+    ///
+    /// [`Bag`]: struct.Bag.html
+    /// [`flush`]: fn.flush.html
     pub unsafe fn defer<R, F: FnOnce() -> R + Send>(&self, f: F) {
         self.defer_garbage(Garbage::new(|| drop(f())))
     }
--- a/src/sync/list.rs
+++ b/src/sync/list.rs
@@ -56,7 +56,9 @@ impl<T> Node<T> {
     pub fn get(&self) -> &T {
         &self.0.data
     }
+}
 
+impl<T: 'static> Node<T> {
     /// Marks this entry as deleted.
     pub fn delete<'scope>(&self, scope: &Scope) {
         self.0.next.fetch_or(1, Release, scope);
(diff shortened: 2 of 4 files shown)


ghost:
It's still a bit annoying that we have to write:
```rust
scope.defer(move || drop(p.into_owned()));
```
instead of just:
```rust
scope.defer(move || p.into_owned());
```

But this is easy to fix - we could just redefine `defer` so that it accepts any `FnOnce() -> R` and ignores whatever the closure returns:

```rust
pub unsafe fn defer<R, F: FnOnce() -> R + Send>(&self, f: F) {
    self.defer_garbage(Garbage::new(|| drop(f())))
}
```

Thoughts on that?

arthurprs:
Weren't we keeping `defer_drop`?

ghost:
@arthurprs 

Yeah, originally we did. But I thought we could perhaps skip a step here and get rid of `defer_drop` as well. Some discussion about that was also [here](https://github.com/crossbeam-rs/rfcs/issues/11#issuecomment-330960329).

At this point `scope.defer_drop(p)` is just a shorter version of `scope.defer(move || drop(p.destroy()))`.
And even less shorter if we we permit `scope.defer(move || p.destroy())`.

arthurprs:
I think `defer_drop` looks shorter and clearer at the cost of a public function. It's a very small price to pay considering it's the most useful variant. Also the name is nicely self documenting.

`scope.defer_drop(n)`
`scope.defer(move || drop(n.into_owned()))`
`scope.defer(move || n.into_owned()))`

jeehoonkang:
I prefer `defer()`-only interface.  defer-dropping an object is 15 bytes longer than before, but I think that's fine: it doesn't add much cognitive load for users.

I also like the proposal to generalize `defer()` so that `scope.defer(move || n.into_owned()))` works.  But I wonder if we can remove `move` here.

src/sync/list.rs
```diff
@@ -3,6 +3,7 @@
 //! Michael.  High Performance Dynamic Lock-Free Hash Tables and List-Based Sets.  SPAA 2002.
 //! http://dl.acm.org/citation.cfm?id=564870.564881
 
+use std::mem::ManuallyDrop;
```
jeehoonkang: What's the benefit of using `ManuallyDrop` here?

ghost: Hmm, yeah - it's not necessary. I've removed `ManuallyDrop` from `sync::list`.

ghost:
@jeehoonkang 

Yeah, `scope.defer(|| n.into_owned()))` will work as well because `into_owned` takes `Ptr` as `self` (consumes it rather than borrows), so the closure will be a *moving* closure.

But note that the compiler can also easily make the closure a *borrowing* closure because `Ptr` is `Copy`. This will happen if you also attempt to access a non-`Copy` variable from the closure.

For that reason, seeing `scope.defer(|| ...)` always makes me worried. Someone may in the future add another line of code to the closure, but forget to also write `move`, and run into nasty UB problems.

ghost:
@arthurprs 

Regarding the interface, I also prefer having only `defer`.
These two are now equivalent:

```rust
scope.defer_drop(n);
scope.defer(move || n.into_owned());
```

While this indeed is the most common application of `defer`, it's at the same not a very common thing to type (it appears only 3 times in the whole crate), so the benefits are kinda low.

If you still feel strongly about this issue, upvote this comment and I'll leave `defer_drop`. No point in too much bikeshedding - we can always revisit this later after gaining some experience. :)

arthurprs:
Don't mind me, it's fine.

In that proposal defer() will wrap the input closure into another closure in order to drop it's result type? That's zero overhead right?

ghost:
Yeah, it wraps into another closure. It's zero overhead because everything gets inlined (I did a few benchmarks to make sure). LLVM is very good at this kind of optimizations.

arthurprs:
Awesome :)

ghost:
@jeehoonkang Can we merge this?

src/sync/queue.rs
```diff
@@ -22,10 +23,12 @@ pub struct Queue<T> {
 
 #[derive(Debug)]
 struct Node<T> {
-    data: T,
+    data: ManuallyDrop<T>,
```
jeehoonkang: I still don't fully understand the motivation to use `ManuallyDrop` here..  Could you please leave some comments on it?

src/sync/list.rs
```diff
@@ -145,7 +149,10 @@ impl<'scope, T> Iter<'scope, T> {
                 ) {
                     Ok(_) => {
                         unsafe {
-                            self.scope.defer_free(self.curr);
+                            // Deferred drop of `T` is scheduled here.
+                            // This is okay because `.delete()` can be called only if `T: 'static`.
```
jeehoonkang, line 153: I think it is safe even tough `T` is not 'static`.  We defer the drop of the object until the global epoch is progressed so that no thread can no longer access the object, isn't it?

ghost:
@jeehoonkang 

#### Node destruction in queues

If you think about it, a `Node<T>` in a queue only *temporarily* holds a `T`.

When you push a node into the queue, it holds a value of type `T`. Then it gets popped and the value is moved *out* from the node and returned to the user. Finally, this *empty* node gets added to the collector, where it's converted to an owned value and destructed (`drop(p.into_owned())`).

During destruction `data` must not be touched because it's empty at that point. The whole point of `ManuallyDrop` is to prevent the destruction of `data` (because in the end it's just an empty slot).

So `data: T` is not the right type here. A better choice would be `Mutex<Option<T>>`, but that is slower and consumes more memory. And `ManuallyDrop<T>` is a compromise - it cannot model optional values, but it's zero cost and at least prevents the unwanted destructor from running.

Before this PR we used `defer_free`, which would convert the node into a `Vec<Node<T>>` of length 0 and capacity 1, then drop it to deallocate memory, preventing the destructor from running.

#### Node destruction in lists

Lists are different from queues. A `Node<T>` in a list owns the `T` for the entirerety of its life. At no point does `data: T` become an empty slot. So this is the right type here.

Moreover, there's an important difference in destruction. When we do `scope.defer(move || p.into_owned())`, the node will be destructed later, possibly at a very distant point in the future. But, since `data` is of type `T`, destruction of the node will also trigger destruction of its `data`. And since `T::drop()` might access `T`'s internals, we must make sure that `T: 'static` because `T::drop()` must not access its possibly non-`'static` references.

This is why destruction of node in a list needs `T: 'static` and in a queue doesn't.

src/mutator.rs
```diff
@@ -247,29 +247,9 @@ impl Scope {
...
-        self.defer_garbage(Garbage::new_free(ptr.as_raw() as *mut T, 1))
-    }
-
-    /// Deferred destruction and deallocation of heap-allocated object `ptr`.
-    // FIXME(jeehoonkang): `T: 'static` may be too restrictive.
-    pub unsafe fn defer_drop<T: Send + 'static>(&self, ptr: Ptr<T>) {
-        self.defer_garbage(Garbage::new_drop(ptr.as_raw() as *mut T, 1))
-    }
-
     /// Deferred execution of an arbitrary function `f`.
```
Vtec234: It would be good to readd the information from the comment on `defer_free` here. For example:
> This function adds a closure to be executed later into a mutator-local [`Bag`]. When the bag becomes full,
> it is flushed into the globally shared queue of bags.

> If the closure environment (captured variables, code) is unusually large, it is wise to follow up with a call
> to [`flush`] so that it doesn't get stuck waiting in the local bag for a long time.

src/sync/list.rs
```diff
@@ -46,15 +48,17 @@ impl<T> Node<T> {
     /// Returns the data in this entry.
     fn new(data: T) -> Self {
         Node(CachePadded::new(NodeInner {
-            data: data,
+            data,
```
Vtec234, line 51: Is this intentional? Why do we ignore `data`?

Vtec234 (approved):

src/sync/list.rs
ghost: It's not ignored -- this is just the *field init shorthand* feature:
https://www.reddit.com/r/rust/comments/5udcju/field_init_shorthand_stabilization_just_landed/

src/mutator.rs
ghost: Readded.

jeehoonkang (approved):

PR #21, opened 2017-09-26
Collector and Handle

jeehoonkang:
## Detailed designs

Realm is a custom domain of memory reclamation.  Currently we have a single global singleton realm.  This PR allows users to define their own realm.  Example:

```rust
use crossbeam_epoch as epoch;

let realm = epoch::Realm::new();

let handle = epoch::Handle.new(&realm);
let handle.pin(|scope| {
    scope.flush();
});
```

User can also use the "default" realm with the existing interface:

```rust
use crossbeam_epoch::{pin, Scope};

pin(|scope| { // pinning to the default realm
    scope.flush();
});
```

## Motivation

There are two motivations behind custom realms.  First, custom realms enable us to test Crossbeam more easily (cf. #17).  In general, it is cumbersome to test singletons.

Second, low-level libraries require custom realms.  For example, `elfmalloc`, a Rust implementation of memory allocator, internally uses Crossbeam but it doesn't want to store the handle (think: a participant of EBR) in the thread-local storage (cf. https://github.com/crossbeam-rs/rfcs/issues/12). 
 In order to do that, `elfmalloc` should be able to make a custom realm and create handles to them.

## Questions

"Realm" might not be the best name for that.

Now `Handle` has a reference to the global data for EBR (`Realm`).  It will add an extra cost of indirection, which might hurt the performance.


arthurprs:
Naming is hard, but Zone or Sector might be better.

I don't think self.realm.xxx() is going to introduce noticeable overhead, calls to these are rare compared to pins() and defer()

jeehoonkang:
@arthurprs I renamed realm to zone.  I think it reflects the notion better.  Thanks!

ghost:
Is there a particular reason why we have `Handle::new(&zone)` instead of `zone.add_handle()`? The latter seems nicer, doesn't it?

`Handle<'a>` is tied to a lifetime because it contains a reference to the underlying `Zone`. I believe the lifetime will almost always be `'static` except in our tests. Unless I'm mistaken, in the [discussion](https://github.com/crossbeam-rs/rfcs/issues/12#issuecomment-328159427) on elfmalloc @joshlf probably had in mind `Handle` without lifetime that holds an `Arc<Zone>` instead of `&'a Zone`.

I can imagine a scenario where you create a thread pool, create one `Handle<'a>` per thread, store handles into thread locals, and then just forget about the `Zone` because you don't really care about it. But you still have to store it somewhere and make sure it lives longer than all the handles. And this kind of memory management is a non-issue if you just use `Arc`...

On bikeshedding:
* Jemalloc has *arenas* (I like thinking of those as *suballocators*).
* Tokio has *reactors* (thread pools driving I/O).
* Rayon has *pools* (of worker threads). There is an implicit global *pool*, but you can create your own *pools* as well.
* By the same logic, I guess we probably have... *collectors*? They collect garbage/deferred functions onto one big pile (and handles do some buffering), then destroy/execute them from time to time. Just an idea...

joshlf:
> Unless I'm mistaken, in the [discussion](https://github.com/crossbeam-rs/rfcs/issues/12#issuecomment-328159427) on elfmalloc @joshlf probably had in mind `Handle` without lifetime that holds an `Arc<Zone>` instead of `&'a Zone`.

That's correct. Another concern I'd have with this design is: is it possible for this to compose nicely? Let's say that I have another handle-based API which wants to use a `Zone` under the hood. If I have to construct a `Zone` explicitly, then I'll put it in the global data. Let's call it `Global`. Each of _my_ handles will have an `Arc<Global>`. Each handle will also need to have a Crossbeam `Handle` which references the global `Zone`. However, that Crossbeam `Handle` will be a field in the `Handle` type that I'm writing, and will need to hold a reference to a different field of the same struct (namely, a reference to the `Zone` in the `Global` which is referenced by the `Arc` field in the handle). I'm not sure whether it's possible to make those lifetimes work out - currently, none of the lifetimes of the fields of a struct outlive one another.

> By the same logic, I guess we probably have... _collectors_?

For what an outsider's opinion is worth, I like this. `Collector` is a nice name for the type. `GC` works too (more pithy), although it's not usually a singular noun (maybe it stands for a "garbage collector"?).

jeehoonkang:
- I renamed zones and handles with collectors and mutators.  I think they better reflect what they do.
- I added `Collector::add_mutator(&self)`, and removed `Mutator::new()`.

On the comment @joshlf just gave and the one on `Arc<Collector>`, I'll think more on it today.  Thanks!

jeehoonkang:
Here is my proposal for composability.

We split `Mutator` into two: (1) a struct that implements the `CollectorAccessor` trait, which has an accessor to the global data `Collector`; and (2) `MutatorInner`: another struct that has other components including the local bag.

Your `Handle` may have `MutatorInner`, and implements `CollectorAccessor` using its own accessor to its own global data.  For more composability, you may be able to define `HandleInner` for the same spirit.

ghost:
Taking inspiration from @joshlf's [logger example](https://github.com/joshlf/handle-example-rs/blob/master/src/lib.rs) (and also `mpsc` and `chase_lev`), would the following be a good interface for use in elfmalloc?

<details>
 <summary>Click to expand</summary>

```rust
struct Collector { ... } // Note: private type

pub struct Handle { // Send + !Sync
    collector: Arc<Collector>,
    // bag, local_epoch, is_pinned, pin_count...
}

impl Handle {
    pub fn pin<F, T>(&self, f: F) -> T where F: FnOnce(&Scope) -> T { ... }
    pub unsafe fn unprotected<F, T>(&self, f: F) -> T where F: FnOnce(&Scope) -> T { ... }
}

impl Clone for Handle {
    // Registers a new handle and returns it.
    fn clone(&self) -> Self { ... }
}

// Creates a new collector and returns the only handle to it.
pub fn collector() -> Handle { ... }

lazy_static! {
    static ref GLOBAL: Handle = collector();
}

thread_local! {
    static LOCAL: Handle = GLOBAL.clone();
}
```
</details>
<br>

A potential oddity here is that we have a useless global handle (basically a sentinel handle), which makes the list a little one node longer than is necessary. But I don't see this is as a big issue.

> We split `Mutator` into two: (1) a struct that implements the `CollectorAccessor` trait, which has an accessor to the global data `Collector`;

I wonder if we need access to the underlying `Collector` at all. With `mpsc` you have `Sender`s and `Receiver`s (which are similar to our handles), but don't have access to the underlying queue.

Our `Collector` would probably have only a single method `add_handle`, but we can achieve the same thing by making handles clonable (just like you can create multiple `Receiver`s by cloning them).

jeehoonkang:
I'm slightly worried about the use of `Arc` here.  I believe we can handle this issue without relying on reference counting..

> I wonder if we need access to the underlying `Collector` at all. With `mpsc` you have `Sender`s and `Receiver`s (which are similar to our handles), but don't have access to the underlying queue.

In the current implementation, each mutator has a `collector: &'scope Collector` as its field.  I'm proposing (1) defining `MutatorInner` without that `collector` field, and (2) defining `Mutator` as a struct of a collector reference field and a `MutatorInner`.

ghost:
> I'm slightly worried about the use of `Arc` here. I believe we can handle this issue without relying on reference counting.

What's the worry? If you're thinking of performance - well, registration in the linked list is already more expensive than atomic increment/decrement. Besides, this happens very rarely (just to spawn a dozen or so handles). Or is it allocation? But we still have to dynamically allocate nodes for the linked list. Or is there something else I'm missing?

> In the current implementation, each mutator has a `collector: &'scope Collector` as its field.  I'm proposing (1) defining `MutatorInner` without that `collector` field, and (2) defining `Mutator` as a struct of a collector reference field and a `MutatorInner`.

It's not obvious to me yet how this solves @joshlf's problem with satisfying lifetimes. It is true that he would be able access the `MutatorInner`, which isn't tied to a lifetime. But the owning `Mutator<'a>` still has to be stored somewhere, and it has the troublesome lifetime. I think he wanted to completely get rid of lifetimes here and model the ownership using `Arc` or something similar to it.

If you think I misunderstood, could you sketch out an API like I did in the previous comment? That might help a bit. :)

joshlf:
Nit: I think these functions you defined are supposed to have `&mut self` parameters? As in:

```rust
impl Handle {
    pub fn pin<F, T>(f: F) -> T where F: FnOnce(&Scope) -> T { ... }
    pub unsafe fn unprotected<F, T>(f: F) -> T where F: FnOnce(&Scope) -> T { ... }
}
```

should actually be:

```rust
impl Handle {
    pub fn pin<F, T>(&mut self, f: F) -> T where F: FnOnce(&Scope) -> T { ... }
    pub unsafe fn unprotected<F, T>(&mut self, f: F) -> T where F: FnOnce(&Scope) -> T { ... }
}
```

> I'm slightly worried about the use of `Arc` here. I believe we can handle this issue without relying on reference counting..

I agree with @stjepang that there shouldn't be much of a performance hit here. The only extra code execution that `Arc` incurs is when you clone, which should be relatively infrequent. During any other method call on a `Handle`, the performance of following an `Arc` reference is identical to the performance of following a `&` reference.

ghost:
@joshlf Yeah, I forgot about that. Also, `mut` is not necessary here: `&self` is good enough.

joshlf:
Do you not need to be able to modify local garbage lists?

ghost:
We do modify it, but it's inside an `UnsafeCell`. Note that when accessing `LOCAL` you don't get a mutable reference (unless you wrap the thread-local into a `RefCell`, which incurs performance cost) so we have to deal with immutable references anyway.

joshlf:
Ah good point.

jeehoonkang:
@stjepang @joshlf I implemented my idea in this branch: https://github.com/jeehoonkang/crossbeam-epoch/tree/handle-composable . You may be interested in the code that composes a logger handle and a GC handle: https://github.com/jeehoonkang/crossbeam-epoch/blob/handle-composable/src/mutator.rs#L384

jeehoonkang:
It seems tokio people are dealing with the exactly same problem: https://github.com/aturon/tokio-rfcs/blob/tokio-reform/tokio-reform.md#rationale-and-alternatives  Roughly speaking, it seems futures : tokio = elfmalloc : crossbeam.

ghost:
I'd still really like to see all `<'scope>`s removed from the file. Those lifetimes just feel overly verbose (appear 48 times in the file!), distracting (more reading/typing for no increase in clarity), and uncomposable (lifetimes are difficult to get around).

From the user's standpoint, what does `&'scope Scope<'scope>` even mean? We are borrowing a `Scope`, and there is another lifetime tied to it... but  why? The `<'scope>` provides no additional information nor safety to the user. Just `&'scope Scope` is good enough, clear, and simple.

And the following looks kinda silly... :)

```rust
fn insert<'scope>(&'scope self, data: T, scope: &'scope Scope<'scope>) -> Ptr<'scope, Node<T>>
                                                              ~~~~~~ what is this?
```

It makes sense for `Ptr<'scope>` to have a lifetime because we handle pointers *by value*. But `Scope`s are always handled *by reference*. There is no way for the user to obtain a `Scope` at all - they can only have `&Scope` (obtained via `pin` and `unprotected`). And this reference already holds the lifetime of the scope, so there's no need to introduce an additional lifetime.

`Mutator<'scope>` implies that mutators live within a scope. This is true in our tests, but is typically not true in `elfmalloc` nor other crates that use `crossbeam`. It's also confusing because `Mutator<'scope>` sounds like something that has to do with `Scope`, but it doesn't.

Also, note that references are *less* composable than `Arc`s and lifetiems can be hard to satisfy. For example, suppose you want to create a `Collector`, spawn 4 threads, give a handle to each one, and automatically destroy the `Collector` as soon as they're done. How would you do that without blocking the current thread?

```rust
let col = Collector::new();
crossbeam::scope(|s| {
    for _ in 0..4 {
        let handle = col.handle();
        s.spawn(move || ...);
    }

    // The current thread blocks here. Can we avoid this?
});
// `col` gets dropped.
// Finally, the collector is destroyed.
```

But, if the collector is backed by an `Arc`, things are much easier:

```rust
let col = Collector::new();
for _ in 0..4 {
    let handle = col.handle();
    thread::spawn(move || ...);
}
// `col` gets dropped.
// The underlying collector is destroyed as soon as all threads finish.
```

In [your example](https://github.com/jeehoonkang/crossbeam-epoch/blob/handle-composable/src/mutator.rs#L497) it is assumed that the global handle lives forever. But what if you want to automatically destroy it as soon as all threads using it exit? It's going to be painful to satisfy lifetimes then.

Note that there is a strong precedent for using `Arc`s in situations like these:

1. `std::thread::Thread` is basically an `Arc` - [here](https://doc.rust-lang.org/nightly/src/std/thread/mod.rs.html#944-946)
2. In `tokio`, `Core` is backed by an `Arc` - [here](https://docs.rs/tokio-core/0.1.9/src/tokio_core/reactor/mod.rs.html#55) and [here](https://docs.rs/tokio-core/0.1.9/src/tokio_core/reactor/mod.rs.html#107)
3. In `rayon`, `ThreadPool` holds an `Arc` - [here](https://docs.rs/rayon-core/1.2.1/src/rayon_core/thread_pool/mod.rs.html#48)

 As you can see, `Arc`s are perfectly valid tools for situations where you want less strict ownership. Don't get me wrong -- lifetimes are a great feature of Rust, but I think this is not the right use case for them.

Also, do we really need traits here (the `Mutator` trait)? This increases the API surface and doesn't seem particularly useful. Why would `elfmalloc` implement `Mutator` trait for its own types instead of just accessing the inner handle? Seems like this API could be made much simpler.

Finally, we might have went overboard with names - there are collectors, mutators, threads, handles, and now even `Impl`. :) Let's pick two - I like `Collector` and `Handle` (those are like `Core` and `Handle` in tokio).

arthurprs:
Is enabling crossbeam to be no_std within the scope of this?

ghost:
@arthurprs Not yet, but it could be a first step towards `no_std`.

ghost:
@jeehoonkang 

Regarding `Scope<'scope>` vs `&'scope Scope`, I'd be okay with either, since they're equivalent.
If we go with the former, then we should make `Scope<'scope>` a `Copy` type and define `pin` like this:

```rust
impl Handle {
    pub fn pin<F, T>(&self, f: F) -> T where F: for<'scope> FnOnce(Scope<'scope>) -> T { ... }
}
```

This essentially makes `Scope<'scope>` a reference to the pinned scope, just like the old `&'scope Scope`.

jeehoonkang:
I fully agree with @stjepang (https://github.com/crossbeam-rs/crossbeam-epoch/pull/21#issuecomment-332826494).  Thank you for the well-thought argument that I can easily understand.

Now I'm implementing what he proposed in https://github.com/crossbeam-rs/crossbeam-epoch/pull/21#issuecomment-332474851 .  I'm almost done, but a problem remains: `Handle` is not `Sync`, but his proposal has `GLOBAL: Handle` in `lazy_static!`.  It simply doesn't compile.  For that reason, maybe we have to have a public `Collector` type that is `Sync`.

ghost:
>  It simply doesn't compile. For that reason, maybe we have to keep the `Collector` type public.

Oh, right. Yeah, let's make `Collector` public. After all, tokio's `Core` is public as well.

jeehoonkang:
@stjepang @joshlf I updated the branch.  Please take a look :)

src/collector.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// A reference to the global epoch.
+    epoch: Epoch,
+}
+
+impl Collector {
+    pub fn new() -> Self {
+        Self { 0: Arc::new(Global::new()) }
+    }
+
+    pub fn add_handle(&self) -> Handle {
```
ghost: Just nitpicking, but there's some precedent for calling this method just `handle`:

* https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Select.html#method.handle
* https://docs.rs/tokio-core/0.1.9/tokio_core/reactor/struct.Core.html#method.handle

src/handle.rs
```diff
@@ -101,31 +105,35 @@ impl<'scope> Mutator<'scope> {
     /// [`Atomic`]: struct.Atomic.html
     pub fn pin<F, R>(&self, f: F) -> R
     where
-        F: FnOnce(&Scope) -> R,
+        F: for<'scope> FnOnce(Scope<'scope>) -> R,
```
ghost: Take a look at signatures of [`scope` in Rayon](https://docs.rs/rayon/0.8.2/rayon/fn.scope.html) and [`scope` in Crossbeam](https://docs.rs/crossbeam/0.3.0/crossbeam/fn.scope.html). It probably doesn't matter, but let's consider writing our signature in a similar manner:
```rust
pub fn pin<'scope, F, R>(&self, f: F) -> R
where
    F: FnOnce(Scope<'scope>) -> R
```

Also, I don't know how to feel about the change from `&'scope Scope` to `Scope<'scope>`. :)

On one hand, `Scope<'scope>` seems sensible: the motivation for parametrizing over the lifetime is the fact that we now have the `global: &'scope Global` field inside `Scope`. But on the other hand, we still have this `bag: *mut Bag`, which is technically just `bag: &'scope mut Bag` in disguise. So this raw pointer is a missed opportunity for using the lifetime.

Why don't we simplify a bit and define `Scope` as:
```rust
struct Scope {
    handle: *const Handle,
}

// or like this:

struct Scope<'scope> {
    handle: &'scope Handle,
}
```

and then reach for `bag` and `global` through the handle?

By the way, note that Rayon's `Scope<'scope>` has `'scope` only because of the `PhantomData` field (see [here](https://docs.rs/rayon-core/1.2.1/src/rayon_core/scope/mod.rs.html#18)), but is otherwise using raw pointers. And Crossbeam's scoped-thread `Scope` doesn't really need `'scope` - it could be implemented without the lifetime as well.

What do you think?

src/handle.rs
```diff
@@ -1,97 +1,101 @@
...
 
 
-impl<'scope> Mutator<'scope> {
-    pub fn new() -> Self {
-        Mutator {
+impl Handle {
+    /// Number of pinnings after which a handle will collect some global garbage.
+    const PINS_BETWEEN_COLLECT: usize = 128;
+
+    pub fn new(global: &Arc<Global>) -> Self {
```
joshlf: Shouldn't this take a `Collector` instead? The fact that a `Collector` is just a newtype wrapper around a `Global` probably shouldn't be exposed to the user (and I'm not even sure that a user could construct an `Arc<Global>` since `Global` itself isn't public outside of this crate).

src/collector.rs
```diff
@@ -0,0 +1,128 @@
...
+        Handle::new(&self.0)
+    }
+}
+
+impl Deref for Collector {
```
joshlf: Do we ever use this impl? If not, it should probably be removed because it's showing up in the list of implemented traits in the documentation, but isn't useful to a user because `Global` isn't exposed, so they couldn't do anything with it.

joshlf:
Echoing a previous comment by @stjepang , is there a reason that `Collector` needs to be public? It's just a newtype around an `Arc<Global>`, and each `Handle` has an `Arc<Global>`, so maybe the interface would be cleaner (and no less powerful) if the only public type was `Handle`?

Perhaps the answer is that `Handle`s are more heavyweight (e.g., maybe the `Bag` they contain is expensive). In that case, it might at least be worth it to add a function to `Handle` that is a synonym for `Collector::new().add_handle()`. Maybe we could switch `new()` to being such a function, since `Collector::add_handle()` already serves the purpose of creating a handle from a `Collector`.

joshlf:
But generally this looks awesome! I'm really excited to be able to use this in elfmalloc :)

arthurprs:
Can we include 1 or 2 simple tests in this PR as well? I'd go a long way visualizing the end result.

src/handle.rs
ghost: This function should be private or `pub(crate)` - handles are created by cloning or calling `.handle()` on the collector.

src/collector.rs
jeehoonkang: Do you mean, (1) we don't implement `Deref` for `Collector`, and (2) implement `push_bag()`, `collect()`, and other functions in `Global` for `Collector`?

src/handle.rs
jeehoonkang: I made it `pub(crate)`.  @joshlf now I think it will not be exposed in the crate doc.  Do you think it is okay?

src/handle.rs
jeehoonkang: - Type signature of `pin()`: After changing the type as you described, I couldn't satisfy the compiler for the following line: https://github.com/jeehoonkang/crossbeam-epoch/blob/handle/src/default.rs#L27 .  What do you think?

- Having `handle` instead of `global` and `bag` in `Scope`: I like the idea.  It should have been implemented in that way from the beginning :)

  But interestingly, the current branch may be better performance-wise, because its access pattern is more shallow than your proposal.  In your proposal, `global` should be accessed **after** `handle` is resolved.  On the other hand, the current branch doesn't have such a dependency.

  I think, however, its hypothetical performance benefit is far less than the benefit of the clarity of your proposal.  I'll change the PR now.

- Type definition of `Scope`: I prefer `handle: &'scope Handle`, because, as you said, it better uses the lifetime logic.  (I actually used `Option<&'scope Handle>` in order to support `fn unprotected()`.  I checked that it is optimized to have the same size with `usize`.

jeehoonkang (commented):
@arthurprs I'd like to add tests in #17.  In the meantime, please look at the example code in a comment: https://github.com/jeehoonkang/crossbeam-epoch/blob/handle/src/collector.rs#L22

@joshlf The primary reason we need `Collector` in addition to `Handle` is, we want a `Sync` reference to the garbage collector, as in https://github.com/jeehoonkang/crossbeam-epoch/blob/handle/src/default.rs#L12 .  As @stjepang suggested, I made `Handle::new()` invisible outside of crate (`pub(crate)`).

I revised the PR after reading the comments.  Thank you for the comments so far!

src/collector.rs
ghost: That's the idea, yeah. But I don't think we need expose `push_bag` in `Collector`. `push_bag` is just an implementation detail, not a publicly visible function.

`collect` would be fine to expose, though, but we should give more thought to the interface. Here's a quick suggestion what a public `collect` function might look like:

```rust
fn collect() -> Collect {
    // ...
}

enum Collect {
    Empty, // No more garbage to collect.
    HasMore, // There is more garbage. You may call `collect()` again if you want.
    BlockedByPin, // There is more garbage, but it can't be collected at the moment.
}
```

src/handle.rs
joshlf: Yep, looks good!

ghost:
@jeehoonkang 

> Type signature of `pin()`: After changing the type as you described, I couldn't satisfy the compiler for the following line: https://github.com/jeehoonkang/crossbeam-epoch/blob/handle/src/default.rs#L27 .  What do you think?

Did you change the signatures of both `pin`s (the one in `impl<'scope> Scope<'scope> { ... }` and the one in `default.rs`)?

src/collector.rs
jeehoonkang: I pushed a commit on the new comments.  In retrospect, it would have been better to put this PR as an RFC first.  If we still has unresolved issues, I'd like to make an RFC based on this branch..

- I made `Collector::collect()` unsafe, because you need to ensure that no concurrent handle is accessing the objects in the garbage queue.

- For the return type of `collect()`, I'd like to make it a separate PR or RFC, as this PR is growing larger and larger.

jeehoonkang:
@stjepang Changing the type of both `Handle::pin()` and `pin()` doesn't work.  But interestingly, changing only the type of `Handle::pin()` works: https://github.com/jeehoonkang/crossbeam-epoch/tree/handle-scope .  I really don't understand the lifetime logic here, though..  If `pin()` anyway needs `for<'scope>`, than I'd like to use it for `Handle::pin()`, too.  If there is another neat solution to this type checking issue, I'd like to adopt it.

Vtec234:
Hey all, I'm really impressed with this work, looks like a good direction for the future. I now see why we need this, but it took me quite a bit of digging around the repo to understand it.

>  In retrospect, it would have been better to put this PR as an RFC first. If we still has unresolved issues, I'd like to make an RFC based on this branch..

For this reason, I agree with the above comment. We've had RFCs for much smaller PRs and this one is a really big change with sizable tradeoffs. I suggest that this should not be merged before at least an overview RFC is accepted. It should contain the motivation behind the changes (step towards supporting no_std and customized allocation/thread environments), the intended usage (consumer such as elfmalloc can provide the same globals that were previously provided by crossbeam) and some of the design details (`Arc` vs. borrows). crossbeam-rs/rfcs#10 contained a bit on this topic, but there's much more in this thread.

Furthermore, we desperately need benchmarks. Without them, it's just guesswork as to what might be more performant. I tried benchmarking this branch on an ARM server and it seemed okay, but the spread (measurement uncertainty) was way too high to call the results meaningful. I'll try to look into this some more. @martinhath also seems to be interested in benchmarking memory recollection schemes.

jeehoonkang:
I wrote an RFC: https://github.com/crossbeam-rs/rfcs/pull/15 .  As @Vtec234 suggested, let's block this PR until an RFC is passed.

src/atomic.rs
```diff
@@ -90,7 +90,7 @@ fn data_with_tag<T>(data: usize, tag: usize) -> usize {
 /// least significant bits of the address.  More precisely, a tag should be less than `(1 <<
 /// mem::align_of::<T>().trailing_zeros())`.
 ///
-/// Any method that loads the pointer must be passed a reference to a [`Scope`].
+/// Any method that loads the pointer must be passed a [`Scope`].
```
ghost: I think this should still be "reference to a [\`Scope\`]"

src/atomic.rs
```diff
@@ -204,7 +204,7 @@ impl<T> Atomic<T> {
     ///     let p = a.load(SeqCst, scope);
     /// });
     /// ```
-    pub fn load<'scope>(&self, ord: Ordering, _: &'scope Scope) -> Ptr<'scope, T> {
+    pub fn load<'scope>(&'scope self, ord: Ordering, _: &'scope Scope) -> Ptr<'scope, T> {
```
ghost: Why do we have a `&'scope self` in all these methods?

src/collector.rs
```diff
@@ -0,0 +1,81 @@
+/// General-purpose garbage collector.
```
ghost: "General-purpose" appears several times in this file. Usually I think of general-purpose GCs as those that are integrated into programming languages (e.g. JVM's GC). Perhaps this should read as "Epoch-based garbage collector"?

src/epoch.rs
```diff
@@ -29,41 +29,45 @@ impl Epoch {
...
     ///
     /// Returns the current global epoch.
     ///
     /// `try_advance()` is annotated `#[cold]` because it is rarely called.
     #[cold]
-    pub fn try_advance<'scope>(&self, registries: &List<LocalEpoch>, scope: &Scope) -> usize {
+    pub fn try_advance<'scope>(
+        &'scope self,
+        registries: &'scope List<LocalEpoch>,
+        scope: &'scope Scope,
```
ghost: Why are all three lifetimes bound like this?

Note that the lifetimes here are covariant so this doesn't affect anything - any `&'scope Foo` reference is a subtype of a reference with an even smaller lifetime: `&'very_small Foo`. So whether you bind these lifetimes together or don't, it doesn't matter.

src/epoch.rs
```diff
@@ -29,41 +29,45 @@ impl Epoch {
...
+    ) -> usize {
         let epoch = self.epoch.load(Relaxed);
         ::std::sync::atomic::fence(SeqCst);
 
-        // Traverse the linked list of mutator registries.
+        // Traverse the linked list of participant registries.
```
ghost: Nit: We could address this later, but -- a *registry* is a place where *registers* or *records* are kept.

More concretely, our linked list is the *registry*, and we usually refer to our *records* in the *registry* (or *entries* in the linked list) as *participants*. :)

The `for registry in registries.iter(scope)` loop would read better like this:
`for participant in registry.iter(scope)`

src/internal.rs
```diff
@@ -0,0 +1,352 @@
...
+        f(&scope)
+    }
+
+    /// Returns `true` if the current participant is pinned.
+    #[inline]
+    pub fn is_pinned(&self) -> bool {
+        self.is_pinned.get()
+    }
+
+    /// Dettaches itself from the garbage collector.
```
ghost: Nit: s/dettaches/detaches
Or just "unregisters".

src/lib.rs
```diff
@@ -8,20 +8,31 @@
...
 //!
 //! There is a global shared instance of garbage queue, which can deallocate ([`defer_free`]) or
 //! drop ([`defer_drop`]) objects, or even run arbitrary destruction procedures ([`defer`]).
 //!
+//! # APIs
+//!
+//! For majority of use cases, just use the default garbage collector by invoking [`pin`]. If you
+//! want to create your own garbage collector, use the [`Collector`] API. When you want to embed a
+//! garbage collector in another systems library, e.g. memory allocator or thread manager, use the
+//! internal API (see [`Global`] and [`Participant`]).
```
ghost: Should this sentence be removed?

src/sync/list.rs
```diff
@@ -59,7 +61,7 @@ impl<T> Node<T> {
 
 impl<T: 'static> Node<T> {
     /// Marks this entry as deleted.
-    pub fn delete<'scope>(&self, scope: &Scope) {
+    pub fn delete<'scope>(&'scope self, scope: &'scope Scope) {
```
ghost: Again, I don't think these arguments need to be bound.

src/sync/queue.rs
```diff
@@ -65,7 +65,12 @@ impl<T> Queue<T> {
     /// Attempts to atomically place `n` into the `next` pointer of `onto`, and returns `true` on
     /// success. The queue's `tail` pointer may be updated.
     #[inline(always)]
-    fn push_internal(&self, onto: Ptr<Node<T>>, new: Ptr<Node<T>>, scope: &Scope) -> bool {
+    fn push_internal<'scope>(
+        &'scope self,
+        onto: Ptr<Node<T>>,
+        new: Ptr<Node<T>>,
+        scope: &'scope Scope,
+    ) -> bool {
```
ghost: Same in this file: `&'scope self`

ghost (changes requested):
One more round of review. Mostly nits.

jeehoonkang:
@stjepang I addressed all your helpful comments.  Thanks!

ghost (approved):
There are still a few more remaining `&'scope self`s, but other than that this looks great. Approving.

Let's wait for one more approval (maybe from @Vtec234?) and we're done.

jeehoonkang:
@stjepang I'm eradicating them now.. :)  It's done.

src/collector.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    /// Collect several bags from the global garbage queue and destroy their objects.
+    ///
+    /// # Safety
+    ///
+    /// It is assumed that no handles are concurrently accessing objects in the global garbage
+    /// queue. Otherwise, the behavior is undefined.
+    #[inline]
+    pub unsafe fn collect(&self) {
+        unprotected(|scope| self.0.collect(scope))
```
Vtec234: Could you give an example scenario where this fails/crashes? Looking at the code in `Global::collect`, it seems safe. The function collects garbage in accordance with the epoch invariant of two epochs having passed since a bag was added to the queue, so there is no way that any handle is concurrently accessing any of the objects that can be deleted (are old enough). The difference between this and calling `Global::collect` inside `pin` is that in the case of `pin`, `try_advance` will never go beyond the current epoch, while here it might, but that's fine, because the thread calling `Collector::collect` is not holding any pointers from the current epoch (unless `Collector::collect` is called in an `unprotected` scope, but that's the user's fault for using a non-exclusive pointer in `unprotected`) and if any other thread/handle has such pointers, it will be pinned.

jeehoonkang: I think the unsafety comes from the fact that `collect()` `defer_drop(ptr)`s (or `defer(move || { ptr.into_owned()})`s in the recent terminology) a pointer `ptr` to a node in the global garbage queue.

Suppose two threads are concurrently calling `collect()`, and a thread just deallocated a node in the garbage queue.  The node is immediately deallocated since we are calling `Global::collect()` within a `unprotected()` scope.  However, the other thread may try to read from a deallocated object, invoking undefined behavior.

I spent some time to come up with this explanation, which basically means it's a documentation bug.  I updated the function doc.  Thank you for pointing out this!

Vtec234: Hah, that's pretty well-hidden unsafety! Thanks for explaining this, it's interesting to see how it propagates down the call stack. On the function comment, since the garbage queue is an implementation detail and probably shouldn't be mentioned in user-facing documentation, how about simply stating instead that:

> It is assumed that no other thread is concurrently using this collector or any of its handles. Otherwise, the behavior is undefined and might lead to memory corruption.

and then explaining the following in a comment inside the function body:

> This is unsafe, because the calling thread may deallocate a node in the global garbage queue immediately due to the use of `unprotected`,  and any other thread that calls `Global::collect()` can then read uninitialized memory while walking through the queue.

Finally, it would be nice to rename the function to `Collector::collect_exclusive()` to make it more obvious.

ghost: So unsafety stems from the fact that we actually don't pin here (we have `unprotected` instead `pin`). And the reason why we don't pin is that there isn't a good way to. We could create a one-off handle to safely collect garbage:

```rust
self.handle().pin(|scope| self.0.collect(scope));
```

but allocating, registering and unregistering a handle just to collect some garbage is obviously not a great solution. I think the right thing to do here would be to implement a special kind of *anonymous pinning* - pinning without a handle (`Collector::pin`). The `try_advance` function would then also check whether anyone has been anonymously pinned.

jeehoonkang: @stjepang I thought about your proposal.  I think the following `Collector::pin()` would be correct:

```rust
// Now a `Global` has the field `anonymous_pin_counter: AtomicUsize`

fn Collector::pin<F, R>(&self, f: F) -> R where ... {
    fetch_add(self.0.anonymous_pin_counter, 1, Relaxed);
    fence(SeqCst);
    let scope = Scope {
        global: &self.0,
        bag: Bag::new()
    };

    defer! {
        fetch_add(self.0.anonymous_pin_counter, -1, Relaxed);
    }

    let result = f(&scope);
    self.0.push_bag(&mut scope.bag, &scope);
    result
}

fn try_advance(...) {
  'L40: let G = EPOCH.load(Relaxed);
  'L41: atomic::fence(SeqCst);

  'LADD: if (anonymous_pin_counter.load(Relaxed) > 0) return;   // ADDED!
  'L42: for th in threads {
  'L43:   let e = th.epoch.load(Relaxed);
  'L44:   if e != SENTINEL && e != G { return; }
  'L45: }
  'L46: atomic::fence(Acquire);

  'L47: EPOCH.store(G + 1, Release);

  'L48: // now objects marked with `<= e - 1` can be deallocated.
}
```

for the [same reason as the handles](https://github.com/crossbeam-rs/rfcs/issues/16).  (Actually it was missing in my [previous proof of Crossbeam](https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-07-23-relaxed-memory.md)!).  Now it's enough for `try_advance()` to check if `anonymous_pin_counter` is zero.


Though this is correct (at least in my opinion), it will block the epoch advance more than necessary.  Once a collector is anonymously pinned, the global epoch can't advance at all.  On the other hand, if we create a one-off handle, the global epoch can advance by one.  I cannot estimate the performance impact of this diff precisely, but I think it can quite bad at least for some use cases.  I'm not sure this "blocking" nature can easily be lifted.


So we have at least three candidate implementation ideas: (1) the unsafe version in the RFC, (2) safe "blocking" version presented above, and (3) hypothetical safe nonblocking version we will think of.  Since it has non-negligible tradeoffs to strike, I'd like to postpone the discussion on `Collector::collect()` to a later RFC, and remove it from the current RFC/PR.

Vtec234 (approved):
Let's just remove `Collector::collect()`, otherwise looks great.

src/collector.rs
ghost: @jeehoonkang Yeah, that seems correct.

Regarding the blocking nature of this algorithm, maybe it's easy to fix after all... :)

Instead of having one counter, let's have two:
```rust
anon_count_even: AtomicUsize,
anon_count_odd: AtomicUsize,
```

To anonymously pin a thread, first load `EPOCH`. Then, if `epoch & 2 == 0` increment `anon_count_even`, otherwise increment `anon_count_odd`.

In `try_advance` just check the counter of opposite parity:
```rust
if G & 2 == 0 && anon_count_odd.load(Relaxed) > 0 { return }
if G & 2 == 1 && anon_count_even.load(Relaxed) > 0 { return }
```

Using this scheme the global epoch can still make progress even in the presence of anonymous pins. Proof of correctness might be a bit tricky, but I believe this should work. :)

jeehoonkang: I like your idea :)  Very neat!

jeehoonkang:
I just removed `Collector::collect()`.  Shall we merge..?

ghost (approved):

src/collector.rs
jeehoonkang: ... now I wonder if why we need to maintain the registry as a participant list.  @stjepang suggested that two `AtomicUsize` are enough for supporting anonymous handles.  If it's true, why don't we extend this idea further to normal handles?

ghost: Because that would create too much contention. If you have a lot of threads modifying two atomic numbers, they will constantly invalidate each other's cache. I did some measurements with this method last year, and the contention would sometimes severely affect performance.

With a linked list every thread modifies its own atomic number so there is no cache invalidation.

jeehoonkang: Beyond performance reasons, I doubt the correctness of the parity method (at least in C/C++ standard).  (For the same reason, I doubt the correctness of the proposed implementation of `Collector::pin()`).  Maybe we can adjust the proof eventually, but for now I couldn't manage to adjust the existing Crossbeam proof for the version with two atomic numbers..

The key reason is, if `pin()` reads a very old stale value of the global epoch, then the pinning thread may be able to read an object, while a concurrent thread is collecting the object.  The parity is not precise abstraction of the local epochs.

PR #26, opened 2017-10-15
Implement Handle::clone()

### LINKED ISSUE #24: Handle should implement Clone
Without implementing `Clone`, users who wish to use a crossbeam `Handle` as part of their own handle-based design will need to ensure that they retain access to the original `Collector` so they can call `.handle()` on it to get a new `Handle`. With `Clone`, a user's handle-based implementation could be as simple as:

```rust
#[derive(Clone)]
struct MyHandle {
    gc: crossbeam_epoch::Handle,
}
```

jeehoonkang:
Closes #24

### DIFF
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -63,6 +63,14 @@ impl Handle {
     }
 }
 
+impl Clone for Handle {
+    fn clone(&self) -> Self {
+        Self::new(self.global.clone())
+    }
+}
+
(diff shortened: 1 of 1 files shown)


ghost (approved):

jeehoonkang:
If no one objects to this PR, I'd like to merge it in a few days.  Thanks!

PR #27, opened 2017-10-15
Fix error in sync mod's trait bound

jeehoonkang:
`Queue<T>` should require `T: Send`, and `List<T>` should require `T: Sync`.

### DIFF
--- a/src/sync/list.rs
+++ b/src/sync/list.rs
@@ -10,7 +10,7 @@ use crossbeam_utils::cache_padded::CachePadded;
 
 
 /// An entry in the linked list.
-struct NodeInner<T> {
+struct NodeInner<T: Sync> {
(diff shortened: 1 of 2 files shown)


ghost:
Why require bounds in structs rather than impls?

Consider an `Arc<T>` - even though it's specifically designed for synchronization, it does not require `T: Sync` or `T: Send`. You can use it in purely single-threaded settings as well. For example, the following works just fine:

```rust
let a = Arc::new(Cell::new(0));
a.set(5);
```

The same applies to `Mutex` and other synchronization primitives. While it doesn't matter much in this situation, it's nice to be able to use concurrent data structures non-concurrently as well.

ghost:
Can we close this? :)

PR #28, opened 2017-10-16
Implement default_handle()

### LINKED ISSUE #25: Provide functions to get the global Collector
If code accepts a `Handle` or a `Collector` that the user is expected to pass, it'd be nice for them to be able to opt out of having to manage their own GC by simply doing `crossbeam::global_collector()` or `crossbeam::global_collector().handle()` or something similar.

jeehoonkang:
Closes #25 

`default_handle()` is unsafe because when it is called in another TLS object's `drop()`, `HANDLE` may have already been dropped.

### DIFF
--- a/src/default.rs
+++ b/src/default.rs
@@ -17,6 +17,16 @@ thread_local! {
     static HANDLE: Handle = COLLECTOR.handle();
 }
 
+/// Returns the default handle.
+///
+/// # Safety
+///
+/// It should not be called in another TLS storage's `drop()`, because `HANDLE` may already be
+/// droped.
+pub unsafe fn default_handle() -> &'static Handle {
+    &*HANDLE.with(|handle| handle as *const _)
+}
+
 /// Pin the current thread.
 pub fn pin<F, R>(f: F) -> R
 where
@@ -32,4 +42,15 @@ pub fn is_pinned() -> bool {
     // FIXME(jeehoonkang): thread-local storage may be destructed at the time `pin()` is called. For
     // that case, we should use `HANDLE.try_with()` instead.
     HANDLE.with(|handle| handle.is_pinned())
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn default_handle_flush() {
+        unsafe { default_handle().pin(|scope| { scope.flush(); }) }
+    }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,5 +74,5 @@ mod sync;
 
 pub use self::atomic::{Atomic, CompareAndSetOrdering, Owned, Ptr};
 pub use self::scope::{Scope, unprotected};
-pub use self::default::{pin, is_pinned};
+pub use self::default::{default_handle, pin, is_pinned};
 pub use self::collector::{Collector, Handle};


joshlf:
Ah, good point. Maybe we should make it `default_handle() -> Option<Handle>` so it can be safe? That basically converts a crash condition into a `None`, and code that knows it won't be called from TLS can safely just `.unwrap()`.

src/default.rs
```diff
@@ -17,6 +17,16 @@ thread_local! {
...
 }
 
+/// Returns the default handle.
+///
+/// # Safety
+///
+/// It should not be called in another TLS storage's `drop()`, because `HANDLE` may already be
+/// droped.
+pub unsafe fn default_handle() -> &'static Handle {
+    &*HANDLE.with(|handle| handle as *const _)
```
Vtec234, line 27: The documentation for `with` is a bit ambiguous, but looking at [the source](https://doc.rust-lang.org/src/std/thread/local.rs.html#277-281) it should always panic if the key has been destroyed. Since we return a reference, it can't escape its scope and so `HANDLE` won't be destroyed while it exists. Hence, this _seems_ safe (but I could be wrong of course).

ghost:
What's the benefit of having a `default_handle()` function over just making `HANDLE` public?

```rust
thread_local! { pub static HANDLE: Handle = COLLECTOR.handle(); }
```

src/default.rs
ghost: It's unsafe because this function allows you to obtain a `&'static Handle` reference and then use it even after `HANDLE` gets destroyed, without any additional checks.

joshlf: @Vtec234 Destruction in the context of TLS is somewhat different than destruction in the rest of Rust unfortunately. When a thread exits, all of the registered TLS variables get destroyed - it doesn't have to do with scope. The problem arises when the `Drop` implementation for a TLS variable accesses another TLS variable which has already been dropped. @ezrosent and I have run into this issue, for example, when our own allocator caches (stored in TLS) were dropped after Crossbeam's TLS was dropped, but we relied on Crossbeam in order to implement `Drop`.

joshlf:
@stjepang IMO, exposing `HANDLE` directly is brittle and makes for a more complicated interface. Users shouldn't have to go read the TLS documentation (which is pretty complicated) just to get a reference to the local handle. Exposing it is also brittle in the sense that it would preclude future changes to the internal implementation.

src/default.rs
Vtec234: In that case another option would be to forward `with` and expose something like `with_default_handle<F: FnOnce(&Handle) -> R>(f: F) -> R`.

joshlf: That's true, but that might lead to callback spaghetti. It's just as easy to do `default_handle() -> Option<&Handle>`, and somewhat more ergonomic.

ghost: @joshlf But how would that work?

Handle is defined like this:

```rust
pub struct Handle {
    global: Arc<Global>,
    local: Local,
}
```

How would you return an `Option<Handle>`? You can't really clone it.

Vtec234: Either `Option<Handle>` or `Handle` (since `with` panics on destroyed TLS, `Option` would probably always be `Some`) can be returned by cloning (#26) the default handle, but that of course has an associated overhead.
@joshlf Yeah, it's not ideal, but it seems to be the only pattern that allows us to restrict a lifetime in the required way.

ghost: @Vtec234 I believe the intention is to return the current thread's handle, not a new one. Cloning registers a brand new handle in the collector.

joshlf: Sorry, I meant to write `Option<&Handle>` instead of `Option<Handle>`. I've updated my original comment.

joshlf: And what's the issue with lifetimes when returning `Option<&Handle>`? Or is the lifetime problem only an issue when returning `Option<Handle>` (ie, not a reference)?

Vtec234: @joshlf: Without a lifetime-binding argument (like in `with`) the lifetime of `&Handle` will have to be `'static`, and that can in turn be leaked until after the TLS drops since Rust allows passing `'static` references around freely.

joshlf: Ah I see. Fair enough.

jeehoonkang: A problem with `fn default_handle() -> Option<&Handle>` is we cannot implement it in Rust stable.  I think we need to use [`HANDLE.try_with()`](https://doc.rust-lang.org/std/thread/struct.LocalKey.html#method.try_with), which is provided only in Rust nightly.

joshlf: Ah good point. Would you guys be cool with just having it be feature-gated on nightly?

jeehoonkang:
Closing (in favor of #31)

Issue #23, opened 2017-10-12
Should Atomic implement Clone?


ghost:
I was implementing a simple Bw-Tree and had a vector of the following type:
```rust
Vec<(K, Atomic<Page<K, V>>)>
```

This vector had to be cloned, but it didn't implement `Clone` because `Atomic<T>` doesn't implement it. So I have to manually go through each element of the vector, load the `Atomic<Page<K, V>>` and then create a clone myself. I should also note that those `Atomic`s never change - they are in fact immutable.

We have this pattern of immutable `Atomic`s actually coming up pretty often. For example, in a lock-free stack nodes are defined as:

```rust
struct Node<T> {
    value: ManuallyDrop<T>,
    next: Atomic<Node<T>>,
}
```

The `next` field becomes immutable once we push the new node onto the stack. This is why loading `next` pointers using `Relaxed` is okay.

Is it a reasonable idea to implement `Clone` for `Atomic<T>`? If so, how would we implement it? Would we load the value using `Relaxed` or maybe `SeqCst`?

jeehoonkang:
I think it's reasonable to implement `Clone` for `Atomic<T>`.  As you said, it's just an immutable pointer to a shared object with interior mutability.  Also I think we don't need an RFC for this change.


> The `next` field becomes immutable once we push the new node onto the stack. This is why loading `next` pointers using `Relaxed` is okay.

Hmm, if I'm understanding your sentence correctly, I disagree with you here..  In order to safely access the data in the next block, loading from the `next` pointer should be ~~`Release`~~ `Acquire`.

arthurprs:
> Hmm, if I'm understanding your sentence correctly, I disagree with you here.. In order to safely access the data in the next block, loading from the next pointer should be Release.

After the node is created, wouldn't `stack.top.cas(old_top, new_top, XXXRelease)` be enough?

jeehoonkang:
@arthurprs Sorry!  I updated the comment.  I meant loading from `next` should be `Acquire`.

ghost:
> I think it's reasonable to implement `Clone` for `Atomic<T>`. As you said, it's just an immutable pointer to a shared object with interior mutability.

[...]

PR #29, opened 2017-10-25
Additional impls and methods in the atomic API

### LINKED ISSUE #22: More impls for Ptr
Since `Ptr` is just a pointer, it might make sense to implement `Eq` for it.

Currently, to compare whether two `Ptr`s are equal, one has to do: `a.as_raw() == b.as_raw()`. Why shouldn't we be able to do just `a == b`?

Maybe we could implement `Ord`, `Hash`, and `std::fmt::Pointer` as well.

ghost:
Closes #22 and #23.

Additions:

* Better `Debug` impls.
* Comparison functions for `Ptr`.
* `Atomic::from_raw`, which is similar to `Owned::from_raw` and `Ptr::from_raw`.
* `Owned::into_box`.

The last one is useful when you need to "abandon" an `Owned` and retrieve the value that is now on the heap. For example:

```rust
let o = Owned::new("foo".to_string());
// ...
let s: String = *o.into_box();
```

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -1,4 +1,6 @@
 use std::borrow::{Borrow, BorrowMut};
+use std::cmp;
+use std::fmt;
 use std::marker::PhantomData;
 use std::mem;
 use std::ops::{Deref, DerefMut};
@@ -93,7 +95,6 @@ fn data_with_tag<T>(data: usize, tag: usize) -> usize {
 /// Any method that loads the pointer must be passed a reference to a [`Scope`].
 ///
 /// [`Scope`]: struct.Scope.html
-#[derive(Debug)]
 pub struct Atomic<T> {
     data: AtomicUsize,
     _marker: PhantomData<*mut T>,
@@ -184,6 +185,20 @@ impl<T> Atomic<T> {
     /// ```
     pub fn from_ptr(ptr: Ptr<T>) -> Self {
         Self::from_data(ptr.data)
+    }
+
+    /// Returns a new atomic pointer pointing to `raw`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::ptr;
+    /// use crossbeam_epoch::{Atomic, Ptr};
+    ///
+    /// let a = Atomic::from_raw(ptr::null::<i32>());
+    /// ```
+    pub fn from_raw(raw: *const T) -> Self {
+        Self::from_data(raw as usize)
     }
 
     /// Loads a `Ptr` from the atomic pointer.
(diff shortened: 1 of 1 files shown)


src/atomic.rs
```diff
@@ -560,6 +575,39 @@ impl<T> Atomic<T> {
     }
 }
 
+impl<T> fmt::Debug for Atomic<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let data = self.data.load(Ordering::SeqCst);
+        let raw = (data & !low_bits::<T>()) as *const T;
```
jeehoonkang: There are several occurrence of these phrases.  Can we make it a function?

src/atomic.rs
```diff
@@ -560,6 +575,39 @@ impl<T> Atomic<T> {
...
+    }
+}
+
+impl<T> Clone for Atomic<T> {
+    fn clone(&self) -> Self {
+        // NOTE: If you're cloning `Atomic`s, then you're *probably* using them as shared immutable
+        // pointers on the heap (i.e. they won't be concurrently modified at the same time they're
+        // being cloned). For that reason we'd most likely be okay with a `Relaxed` load here.
+        // However, just to avoid potential subtle pitfalls, `SeqCst` is used instead as the most
+        // conservative and safest choice of ordering.
```
jeehoonkang: I prefer `Relaxed` here, because `clone()` is not necessarily associated with any synchronization.  It's semantics would be copy **a** value of the atomic, but without synchronization it can be a stale one.  If the user wants some synchronization, s/he can do it with other atomic operations or fences.

jeehoonkang (changes requested):
Thank you for the PR!  It's quite comprehensive, I think.  I'd like to discuss the ordering used in `Atomic::clone()`, but except for that I'm happy with all the changes.

src/atomic.rs
ghost: I agree. But then again, `Debug` implementation for `AtomicPtr` uses the `SeqCst` ordering, as you can see [here](https://doc.rust-lang.org/nightly/src/core/sync/atomic.rs.html#1811-1815). It could/should use `Relaxed` by the same reasoning, right? My guess is that `SeqCst` is used "just in case" because it's the safest option.

jeehoonkang: I'm mixed for using `Relaxed` in `Debug`..

 On one hand, I'd like to say we should put `Relaxed` there and requires users to put `fence(SeqCst)` if they are not sure of orderings, since `Debug::fmt()` has nothing to do with synchronization.  The same logic for `clone()`.

On the other hand, `Debug::fmt()` is primarily for debugging purposes, as opposed to `fn clone()` which can be used in production, and for that using `SeqCst` is perfectly okay.  However, the use of `SeqCst` here should not imply that `Debug::fmt()` is always using a "safe" ordering..

ghost: That's fair enough. :) I'll switch to `Relaxed` ordering.

jeehoonkang (approved):

src/atomic.rs
```diff
@@ -560,6 +584,37 @@ impl<T> Atomic<T> {
...
+impl<T> fmt::Pointer for Atomic<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let data = self.data.load(Ordering::SeqCst);
+        let (raw, _) = decompose_data::<T>(data);
+        fmt::Pointer::fmt(&raw, f)
+    }
+}
+
+impl<T> Clone for Atomic<T> {
+    /// Returns a copy of the atomic value.
```
Vtec234, line 608: Rust docs don't show comments on trait function implementations, so maybe it'd be good to put a sentence in the `Atomic` struct description stating that `clone` and formatting methods are unsychronised.

Vtec234 (approved):

src/atomic.rs
ghost: Actually, documentation on trait methods can be overridden and rustdoc shows it as you would expect. :)

Addressing this peculiarity around `Relaxed` in clones would probably be best explained as a dedicated subsection that explains how to use `Atomic`s when you don't really need atomicity, but just convenient access to heap-allocated objects. I'll address that in a future PR. A big one involving the switch from scopes to guards is coming up anyway...

Vtec234: Hm, I could swear the doc wasn't there, but you're right and I probably just missed it.

PR #31, opened 2017-11-03
Replace Scope with Guard

### LINKED ISSUE #25: Provide functions to get the global Collector

ghost:
Apologies for the huge PR. I'll try to summarize what's going on here:

1. `atomic.rs` - No big changes, `&Scope` was just changed to `&Guard`.
2. `collector.rs` - `Handle::clone` now creates another reference instead of a new handle.
3. `epoch.rs` - Almost everything is new. I've implemented types `Epoch` and `AtomicEpoch`, which are similar to `usize` and `AtomicUsize`.
4. `guard.rs` - No big deal, just a lot of documentation.
5. `internal.rs` - The meatiest part of this PR.
6. `sync/list.rs` - This file is gone. I've found it easier to work directly with the linked list in `internal.rs` rather than through a `List<T>` abstraction. But we could bring `List<T>` back in a future PR, though.
7. `sync/queue.rs` - Scopes replaced with guards, that's all.

Closes #25.


ghost:
@jeehoonkang  Just a reminder that this is waiting for review. :)

src/epoch.rs
```diff
@@ -7,77 +7,116 @@
...
+            data: self.data.wrapping_add(2),
+        }
+    }
 
-    fn deref(&self) -> &Self::Target {
-        &self.epoch
+    /// Decomposes the internal data into the epoch and the pin state.
+    #[inline]
+    fn decompose(&self) -> (usize, bool) {
+        (self.data >> 1, (self.data & 1) == 1)
```
jeehoonkang, line 77: What about returning `(self.data, (self.data & 1) == 1)`?  Maybe it's a little bit more difficult interface to use, but may we can save an instruction :smiley:

src/guard.rs
```diff
@@ -0,0 +1,305 @@
...
+///     }
+/// }
+/// ```
+///
+/// [`Atomic`]: struct.Atomic.html
+/// [`defer`]: struct.Guard.html#method.defer
+#[inline]
+pub unsafe fn unprotected() -> &'static Guard {
+    static UNPROTECTED: usize = 0;
+    &*(&UNPROTECTED as *const _ as *const Guard)
```
jeehoonkang: I couldn't figure out why it is safe -- would you please explain a little bit more?

src/internal.rs
```diff
@@ -61,237 +54,307 @@ impl Global {
...
+    guard_count: Cell<usize>,
+
+    /// The number of active handles.
+    handle_count: Cell<usize>,
+
     /// Total number of pinnings performed.
-    pin_count: Cell<usize>,
+    ///
+    /// This is just an auxilliary counter that sometimes kicks off collection.
+    pin_count: Cell<Wrapping<usize>>,
```
jeehoonkang, line 172: wow, `Wrapping` is very convenient :)

jeehoonkang (commented):
Besides minor comments, I'm happy with all the changes.  Thanks!

But I'm a little bit hesitant to merge this PR now, because the following issue has not been resolved yet: https://github.com/crossbeam-rs/rfcs/issues/23  In order to address the issue, maybe we need to once more change this repo.

What do you think?

src/epoch.rs
ghost: Well, it's just a shift operation that gets executed on cold paths, so it doesn't really matter... :)

src/guard.rs
ghost: Added an explanation.

ghost:
@jeehoonkang Do you think we could merge this now and release version 0.1.0?
I'd like to release a 0.1.0 so that crossbeam-deque and crossbeam-channel can start depending on it. Then I could finally deprecate coco and switch rayon onto crossbeam. :)

Regarding crossbeam-rs/rfcs#23, I think we will have to make slight adjustments to `Collector` or `Guard` either way, but figuring that out will again take some time.

ghost:
Also, I forgot to ask:

Do you have an opinion on whether `unprotected()` should return a `&'static Guard` or a `Guard`? Originally I decided to go with returning a reference, but now I'm not so sure that was the right choice...

jeehoonkang:
@stjepang I also thought of releasing 0.1.0 after merging it, but was also concerned with backward compatibility.  If we release it now, it's quite clear that 0.2.0 will break some API.  That's legal in semantic versioning scheme, but still it's better to avoid breaking changes if we can..

On the other hand, crossbeam-epoch 0.1.0 is so waited in many parties, including crossbeam itself and rayon  If we can provide a clear migration path from 0.1.0 to 0.2.0, breaking changes will not be fatally painful.

I'm torn between the two options, but anyway I'm willing to follow your decision.  Let's merge it!

I'd like to say thank you for your great job!

jeehoonkang:
On `unprotected()`: I'm also not sure of this issue, but I think it's easier to change the API from returning `&'static Guard` to returning `Guard` than the other way around.  How about beginning with a more strict API, and relaxing it later when use cases demand it?

PR #32, opened 2017-11-20
Add simple benchmarks

ghost:
Here's an initial set of simple benchmarks.

### DIFF
--- a/benches/flush.rs
+++ b/benches/flush.rs
@@ -1,0 +1,51 @@
+#![feature(test)]
+
+extern crate crossbeam_epoch as epoch;
+extern crate crossbeam_utils as utils;
+extern crate test;
+
+use std::sync::Barrier;
+
+use test::Bencher;
+use utils::scoped::scope;
+
+#[bench]
+fn single_flush(b: &mut Bencher) {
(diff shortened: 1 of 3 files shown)


benches/flush.rs
```diff
@@ -0,0 +1,51 @@
...
+use std::sync::Barrier;
+
+use test::Bencher;
+use utils::scoped::scope;
+
+#[bench]
+fn single_flush(b: &mut Bencher) {
+    const THREADS: usize = 16;
+
+    let start = Barrier::new(THREADS + 1);
```
jeehoonkang, line 16: I wonder if is it possible that the cost of `Barrier` dominates the latency of `single_flush`.  Do you have an insight on the numbers?

jeehoonkang (approved):
Thanks!

I just left a minor question, but I think even without answering it it's worth merging this PR.

benches/flush.rs
ghost: I've just tried inserting sleeps after `start.wait()` so that barriers don't get touched during the benchmark. It didn't affect the results in any measurable way.

Issue #30, opened 2017-10-25
Removing *_owned methods


ghost:
Have a look at the following methods on `Atomic<T>` we currently implement:

<details>
  <summary>Click to expand</summary>

```rust
impl<T> Atomic<T> {
    pub fn store(&self, new: Ptr<T>, ord: Ordering) {
        self.data.store(new.data, ord);
    }

    pub fn store_owned(&self, new: Owned<T>, ord: Ordering) {
        let data = new.data;
        mem::forget(new);
        self.data.store(data, ord);
    }

    pub fn swap<'scope>(&self, new: Ptr<T>, ord: Ordering, _: &'scope Scope) -> Ptr<'scope, T> {
        Ptr::from_data(self.data.swap(new.data, ord))
    }

    pub fn compare_and_set<'scope, O>(
        &self,
        current: Ptr<T>,
        new: Ptr<T>,
        ord: O,
        _: &'scope Scope,
    ) -> Result<(), Ptr<'scope, T>>
    where
        O: CompareAndSetOrdering,
    {
        match self.data.compare_exchange(
            current.data,
            new.data,
            ord.success(),
            ord.failure(),
        ) {
            Ok(_) => Ok(()),
            Err(previous) => Err(Ptr::from_data(previous)),
        }
    }

    pub fn compare_and_set_owned<'scope, O>(
        &self,
        current: Ptr<T>,
        new: Owned<T>,
        ord: O,
        _: &'scope Scope,
    ) -> Result<Ptr<'scope, T>, (Ptr<'scope, T>, Owned<T>)>
    where
        O: CompareAndSetOrdering,
    {
        match self.data.compare_exchange(
            current.data,
            new.data,
            ord.success(),
            ord.failure(),
        ) {
            Ok(_) => {
                let data = new.data;
                mem::forget(new);
                Ok(Ptr::from_data(data))
            }
            Err(previous) => Err((Ptr::from_data(previous), new)),
        }
    }
}
```
</details><br>

It's interesting that `swap_owned` is missing, although it is not stricly speaking necessary: `a.swap_owned(o, ord, scope)` would be equivalent to `a.swap(o.into_ptr(scope), ord, scope)`. But, for the sake of consistency and convenience, we should probably have `swap_owned` as well.

There's a lot of duplication. We have one set of functions for `Ptr` and another for `Owned`. Let's try to unify them. First, we could introduce a trait that is implemented by `Ptr` and `Owned`:

```rust
[...]

PR #33, opened 2017-11-20
Panic if the pointer is null in into_owned

ghost:
This is a just small quality-of-life improvement. It's better to fail early when converting a null `Ptr` into `Owned` than run into nasty memory issues and potential SIGSEGVs later.

The overhead of testing whether the pointer is null should be really small. The `into_owned` method is typically used to drop the value and deallocate memory, so one quick check is not going to affect performance.

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -1030,7 +1030,7 @@ impl<'g, T> Ptr<'g, T> {
     ///
     /// # Panics
     ///
-    /// Panics if this pointer is null.
+    /// Panics if this pointer is null, but only in debug mode.
     ///
     /// # Safety
     ///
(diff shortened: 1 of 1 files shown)


Vtec234 (approved):

jeehoonkang (approved):

jeehoonkang:
How about asserting only in debug mode? `debug_assert!(...)`

Because in implementing #30, I had to use `Owned::from_data` in `compare_and_set`.  It's in a reasonably hot path.  `Owned` should be converted to `usize` for performing CAS, and then converted back to `Owned` when CAS fails.

ghost:
On another thought, shouldn't `into_owned()` return `Option<Owned<T>>` instead of `Owned<T>`?
That might be even more foolproof.

Vtec234:
@stjepang: that would add a cumbersome `unwrap()` in 99% of the uses where the user expects the `Ptr` to be valid and a slightly nicer API for the 1% when the `Ptr` might actually be null, which can just be checked with `is_null`. `debug_assert` might be better though.

ghost:
Fair enough - I've switched to `debug_assert!`. Merging...

PR #36, opened 2017-11-21
Remove from_* and into_* methods

### LINKED ISSUE #34: Removing `from_*` methods
e.g. as far as we implement `From<Owned<T>>` for `Atomic<T>`, `Atomic::from_owned()` is redundant.  I don't see much benefit of having these methods..

jeehoonkang:
Closes #34 

As discussed in #34, strangely, Rust's type inference engine cannot resolve an obvious type inference problem: https://github.com/jeehoonkang/crossbeam-epoch/commit/6def6e85d42716bd53ba790794336240d9159f5f#diff-422fa30cb418a44869868fb187a6ded0R1030  That renders using `From` less ergonomic.  @stjepang said that it will only occur for toy examples, but I'm not entirely sure of that...  For this reason, I'm not 100% confident for the changes in this PR.

What do you think?

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -946,7 +946,10 @@ impl<'g, T> Shared<'g, T> {
     /// }
     /// ```
     pub unsafe fn into_owned(self) -> Owned<T> {
-        debug_assert!(self.as_raw() != ptr::null(), "converting a null `Shared` into `Owned`");
+        debug_assert!(
+            self.as_raw() != ptr::null(),
+            "converting a null `Shared` into `Owned`"
+        );
         Owned::from_data(self.data)
     }
 
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -116,7 +116,7 @@ unsafe impl<T: Send + Sync> Sync for Atomic<T> {}
 impl<T> Atomic<T> {
     /// Returns a new atomic pointer pointing to the tagged pointer `data`.
     fn from_data(data: usize) -> Self {
-        Atomic {
+        Self {
             data: AtomicUsize::new(data),
             _marker: PhantomData,
         }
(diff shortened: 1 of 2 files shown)


src/atomic.rs
```diff
@@ -115,7 +115,7 @@ unsafe impl<T: Send + Sync> Sync for Atomic<T> {}
 impl<T> Atomic<T> {
     /// Returns a new atomic pointer pointing to the tagged pointer `data`.
     fn from_data(data: usize) -> Self {
-        Atomic {
+        Self {
```
ghost, line 128: I've noticed you really like using the `Self` keyword, and I do too. :)

Just one comment, though... Take a look at the documentation of [with_capacity](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity). I find browsing documentation much easier when types are explicitly spelled out in signatures rather than abbreviated with `Self`.

We should probably avoid `Self` in public signatures, but otherwise use it pretty liberally throughout the rest of the code. The standard library apparently uses the same approach.

ghost (approved):
Looks pretty good to me, but let's see if others agree. :)

src/atomic.rs
jeehoonkang: > I find browsing documentation much easier when types are explicitly spelled out in signatures rather than abbreviated with `Self`.  ...  We should probably avoid `Self` in public signatures, but otherwise use it pretty liberally throughout the rest of the code.

I personally disagree here; if `Self` is more difficult to read in the documentation, maybe it's a bug in `rustdoc`..  But more importantly,

>  The standard library apparently uses the same approach.

I'm fully willing to follow the standard library's approach.  Tomorrow I'll rename `Self` into the actual type names in the public signatures.

ghost: > I personally disagree here; if `Self` is more difficult to read in the documentation, maybe it's a bug in `rustdoc`

Great idea! I'll open an issue on this in the rust repository.

Vtec234:
I suggest going forward with removing these methods and later if the type inference becomes a nuissance, adding them back. That's because doing so is a backwards-compatible change, but removing them later is not.

jeehoonkang:
I rebased this PR and removed `Self` from the public signatures.  I think it's ready to be merged!

PR #37, opened 2017-11-21
Custom type for compare-and-set errors

ghost:
This PR changes the return type of `compare_and_set` so that the error case returns a struct instead of a pair of pointers. The problem with pairs is that it's easy to confuse which pointer is the first and which one is the second.

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -900,14 +900,14 @@ impl<'g, T> Shared<'g, T> {
     ///
     /// Dereferencing a pointer is unsafe because it could be pointing to invalid memory.
     ///
-    /// Another concern is the possiblity of word races due to lack of proper synchronization.
+    /// Another concern is the possiblity of data races due to lack of proper synchronization.
     /// For example, consider the following scenario:
     ///
     /// 1. A thread creates a new object: `a.store(Owned::new(10), Relaxed)`
     /// 2. Another thread reads it: `*a.load(Relaxed, guard).as_ref().unwrap()`
     ///
     /// The problem is that relaxed orderings don't synchronize initialization of the object with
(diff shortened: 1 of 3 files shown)


src/atomic.rs
```diff
@@ -887,14 +900,14 @@ impl<'g, T> Shared<'g, T> {
     ///
     /// Dereferencing a pointer is unsafe because it could be pointing to invalid memory.
     ///
-    /// Another concern is the possiblity of data races due to lack of proper synchronization.
+    /// Another concern is the possiblity of word races due to lack of proper synchronization.
```
Vtec234: Search and replace? ;)

ghost: Haha, you have a good eye!

src/atomic.rs
```diff
@@ -376,12 +388,12 @@ impl<T> Atomic<T> {
         O: CompareAndSetOrdering,
         P: Pointer<T>,
     {
```
Vtec234: Shouldn't weak also return `CasError`?

ghost: Doh, I'm so sloppy! Fixed.

src/atomic.rs
Vtec234: I would stick to `data` though, since `word` is generally used when the specific width of the object matters, and is quite ambiguous. In particular, x86 intel assembly refers to words as being 2 bytes rather than the machine word (4 on x86, 8 on x86_64) size.

ghost: Hm, I didn't know that. Okay, I'm reverting back to `data`. Maybe we'll come up with a better name next time...

Vtec234 (approved):

jeehoonkang (approved):

PR #38, opened 2017-11-22
Remove Send bound from defer

ghost:
Just like `defer` doesn't require `F: 'static` anymore, we don't really need `F: Send` either. These invariants should be ensured by the caller. After all, that is one of the reasons why `defer` is an unsafe function.

I'm also removing the `Send` impl for `Shared`, which basically exists only to satisfy the `F: Send` bound in `defer`.

### DIFF
--- a/src/garbage.rs
+++ b/src/garbage.rs
@@ -39,9 +39,6 @@ pub struct Garbage {
     func: Deferred,
 }
 
-unsafe impl Sync for Garbage {}
-unsafe impl Send for Garbage {}
-
 impl fmt::Debug for Garbage {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
         write!(f, "garbage {{ ... }}")
@@ -50,7 +47,7 @@ impl fmt::Debug for Garbage {
 
 impl Garbage {
     /// Make a closure that will later be called.
-    pub fn new<F: FnOnce() + Send>(f: F) -> Self {
+    pub fn new<F: FnOnce()>(f: F) -> Self {
         Garbage { func: Deferred::new(move || f()) }
     }
 }
--- a/src/garbage.rs
+++ b/src/garbage.rs
@@ -39,6 +39,10 @@ pub struct Garbage {
     func: Deferred,
 }
 
+
+unsafe impl Sync for Garbage {}
+unsafe impl Send for Garbage {}
+
 impl fmt::Debug for Garbage {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
         write!(f, "garbage {{ ... }}")
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -172,8 +172,6 @@ pub struct Local {
     pin_count: Cell<Wrapping<usize>>,
 }
 
-unsafe impl Sync for Local {}
-
 impl Local {
     /// Number of pinnings after which a participant will execute some deferred functions from the
     /// global queue.
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -172,6 +172,8 @@ pub struct Local {
     pin_count: Cell<Wrapping<usize>>,
 }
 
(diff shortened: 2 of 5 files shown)


Vtec234:
Is this a substantial gain for ergonomics and if so, could you give a short example? My understanding was that while `'static` is too much of a burden due to lifetime restrictions, the `Send` trait is easily implementable and so having it present as a bound on `defer` doesn't make it much harder to use, while making an important invariant explicit.

ghost:
My feeling is that even though `F: Send` is a perfectly sensible requirement for `defer`, this trait bound doesn't bring much to the table when it comes to safety and adds to the confusion.

When implementing a concurrent data structure, we ensure that the interface is safe to use from the outside by requiring `T: 'static` where necessary and implementing `Send` and `Sync` manually.

Take `Queue<T>` for example. These are our `Send`/`Sync` requirements from the outside:

```rust
unsafe impl<T: Send> Sync for Queue<T> {}
unsafe impl<T: Send> Send for Queue<T> {}
```

On the inside, `F: Send` on `defer` (when destroying popped nodes) requires us to add lines like the following, otherwise the code doesn't compile:

```rust
unsafe impl<T> Send for Node<T> {}
```

Now, that line is required because of `T: Send` on `Send` impl for `Shared`:

```rust
unsafe impl<'g, T: Send> Send for Shared<'g, T> {}
```

This `Send` impl for `Shared` is already kinda fishy and I don't feel comfortable about it. It literally exists for the sole reason to satisfy the `F: Send` bound on `defer`. Other than that, it doesn't make much sense to `Send` pointers of type `Shared` around.

All that charade makes me think that the `Send` safety on `defer` is totally bogus. In fact, note that there is a mistake in the `Send` impl for `Node<T>`:

```rust
// This should require `impl<T: Send>`.
unsafe impl<T> Send for Node<T> {}
```

This just means the impl was hastily added in order to get around the compilation error. :)

My opinion is that we don't actually get any tangible safety this way - we're just adding confusing impls left and right. So `F: Send` is a viral requirement just like `F: 'static` was, and we have to deal with safety issues outside `defer` anyway.

jeehoonkang:
I agree with @stjepang, except for the discussion on `Shared: Send`; to my eyes, it's intuitive right.  I see that it's currently not used at all, but why do you actually want to remove it?

ghost:
@jeehoonkang Because each `Shared` is associated with a `Guard`, and guards are associated with the current thread. Sending `Shared` pointers to other threads doesn't make much sense, does it?

The only situation in which we want to send a `Shared` into another thread is when moving the pointer into a closure passed to `defer`. But this is a very special case, and `defer` is a magical function that breaks all kinds of rules (it even works around lifetimes).

So I believe `impl Send for Shared` is:

1. Not useful. I don't see why would anyone send a `Shared` into another thread (except through `defer`, of course).
2. Confusing. `Guard` is not `Send`, so why would `Shared` be?

jeehoonkang:
@stjepang theoretically a `Guard` in thread 1 (T1) can be borrowed with lifetime `'g` and T2 may use a `Shared` sent from T1 within the lifetime `'g`..

Yes, that's very much contrived.  Let's remove `Shared: Send`.

jeehoonkang (approved):

src/garbage.rs
```diff
@@ -39,9 +39,6 @@ pub struct Garbage {
     func: Deferred,
 }
 
-unsafe impl Sync for Garbage {}
-unsafe impl Send for Garbage {}
```
Vtec234: I would keep this impl as it encodes the assertion that we need `Garbage` to be sendable across threads.

Vtec234 (approved):

src/internal.rs
```diff
@@ -172,8 +172,6 @@ pub struct Local {
     pin_count: Cell<Wrapping<usize>>,
 }
 
-unsafe impl Sync for Local {}
```
Vtec234: Would also keep this, since we do access `Local` from many threads.

PR #39, opened 2017-11-22
Add a thread sanitization test

ghost:
I've copied a test from coco which aggressively spawns threads and updates/loads/replaces a shared `Atomic`, constantly creating new objects and destroying old ones.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,13 +13,21 @@ categories = ["concurrency", "memory-management"]
 
 [features]
 default = ["use_std"]
+<<<<<<< HEAD
 use_std = ["lazy_static", "crossbeam-utils/use_std"]
+=======
+>>>>>>> Add sanitization tests
 nightly = ["arrayvec/use_union"]
-strict_gc = []
+sanitize = []
+use_std = ["lazy_static"]
 
 [dependencies]
 arrayvec = { version = "0.4", default-features = false }
+cfg-if = "0.1"
 crossbeam-utils = { version = "0.2", default-features = false }
 lazy_static = { version = "0.2", optional = true }
 memoffset = { version = "0.1", default-features = false }
 scopeguard = { version = "0.3", default-features = false }
+
+[dev-dependencies]
+rand = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,13 +13,9 @@ categories = ["concurrency", "memory-management"]
 
 [features]
 default = ["use_std"]
-<<<<<<< HEAD
-use_std = ["lazy_static", "crossbeam-utils/use_std"]
-=======
->>>>>>> Add sanitization tests
 nightly = ["arrayvec/use_union"]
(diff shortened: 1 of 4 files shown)


src/garbage.rs
```diff
@@ -29,11 +29,13 @@ use arrayvec::ArrayVec;
...
 /// Maximum number of objects a bag can contain.
-#[cfg(not(feature = "strict_gc"))]
-const MAX_OBJECTS: usize = 64;
-#[cfg(feature = "strict_gc")]
-const MAX_OBJECTS: usize = 4;
-
+cfg_if! {
+    if #[cfg(feature = "sanitize")] {
+        const MAX_OBJECTS: usize = 4;
+    } else {
```
Vtec234: This removes the `strict_gc` feature, shouldn't it be `feature = sanitize` OR `feature = strict_gc` instead?

Cargo.toml
```diff
@@ -13,10 +13,11 @@ categories = ["concurrency", "memory-management"]
 
 [features]
 nightly = []
-strict_gc = []
+sanitize = []
```
Vtec234: Do we not want two orthogonal `sanitize` and `strict_gc` features? `strict_gc` doesn't add the `Acquire` hack in `internal.rs`.

examples/sanitize.rs
```diff
@@ -0,0 +1,60 @@
...
+            unsafe {
+                p.deref().fetch_add(sum, Relaxed)
+            }
+        };
+
+        sum = sum.wrapping_add(val);
+
+        if rng.gen_range(0, 2000) == 0 {
+            let a = a.clone();
+            thread::spawn(move || worker(a)).join().unwrap();
```
Vtec234: I'm curious, what's the purpose of spawning threads recursively? Could the same effect be achieved by simply spawning more threads in `main` and having the loop in `worker` be tighter (no thread spawn overhead)?

Cargo.toml
jeehoonkang: I think the strictness should be configurable with more flexibility: the tradeoff between latency and peak working set is one of the most important parameters for garbage collection.  But I think we can tentatively remove `strict_gc` support now.  Anyway, the size of a bag seems too coarse-grained a parameter for the strictness.

jeehoonkang (approved):

src/garbage.rs
ghost: I think we're okay with the single feature `sanitize` for now. Let's add `strict_gc` (or maybe call it `aggressive_gc`?) alongside it later when we find a use for it.

examples/sanitize.rs
ghost: It's just to randomize the interleaving of thread spawning/exiting and atomic operations. It's a clumsy implementation, I admit.

I've changed the test now so that threads are spawned in advance and each thread first sleeps for a random amount of time, then works with the atomic for a random amount of time.

PR #41, opened 2017-11-22
Revert the linked list implementation

ghost:
Today I noticed worrying spurious failures on Travis:
https://travis-ci.org/crossbeam-rs/crossbeam-epoch/jobs/305533939

There was another failure (again, `sync::queue::test::push_try_pop_many_mpmc` got stuck) that got "fixed" by restarting the tests.

First I thought that something must be wrong with the queue implementation. Unfortunately, I couldn't reproduce the error on my machine nor find anything suspicious in it.

Then I switched `crossbeam-channel` from `coco` to our latest version of `crossbeam-channel`. One test started failing consistently (you can run it with `cargo test --test list drops`). After a lot of debugging, I figured that calls to `List::insert()` sometimes get stuck. Weird, since there is nothing suspicious in its implementation either. Then I tried replacing all memory orderings in `list.rs` with `SeqCst` and commenting out calls to `defer`. The problem persisted.

Finally, I decided to see what happens if I replace `List<T>` with the old implementation that used `next` pointers directly inside `Local`s. Interestingly, the problem seemed to be gone. After spending some more time looking at `list.rs` and tinkering with it, I couldn't find the bug.

Since we're going towards 0.1.0 and the old list implementation seems to be working fine (it's also basically copy-pasted from coco, which has been working for a while now), I suggest we use that one for now. Moreover, it's probably easier to audit since it contains fewer moving parts (this PR has +86/-375 changes, plus there are no complications around intrusiveness and offsetof).


ghost:
Actually, there's just a silly bug in `List::insert`. My bad!

@Vtec234 will submit a fix.

PR #43, opened 2017-11-24
Implement Guard::safepoint()

jeehoonkang:
This is the implementation of `Guard::safepoint()` we discussed in https://github.com/crossbeam-rs/rfcs/issues/18.

I don't expect it to be merged before 0.1.0, but I'm personally in favor of doing so; I need it in implementing a wait-free queue.

### DIFF
--- a/src/guard.rs
+++ b/src/guard.rs
@@ -194,6 +194,38 @@ impl Guard {
         }
     }
 
+    /// Unpins and then pins the current thread.
+    ///
+    /// Since `&mut self` is given, we can statically deduce that there is no [`Shared`] pointer
+    /// that is created before an invocation of `safepoint()` and then survives after the
+    /// invocation.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_epoch as epoch;
+    ///
+    /// let mut guard = epoch::pin();
+    /// {
+    ///     let a = epoch::Owned::new(42u64).into_shared(&guard);
+    ///     unsafe { guard.defer(move || a.into_owned()) };
+    /// }
+    /// // `a` should not survive.
+    /// guard.safepoint();
+    /// {
+    ///     let b = epoch::Owned::new(42u64).into_shared(&guard);
+    ///     unsafe { guard.defer(move || b.into_owned()) };
+    /// }
+    /// guard.flush();
+    /// ```
+    ///
+    /// [`Shared`]: struct.Shared.html
+    pub fn safepoint(&mut self) {
+        if let Some(local) = unsafe { self.local.as_ref() } {
+            local.repin();
+        }
+    }
+
     /// Temporarily unpins the thread, executes the given function and then re-pins the thread.
     ///
     /// This method is useful when you need to perform a long-running operation (e.g. sleeping)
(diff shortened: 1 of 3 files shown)


src/guard.rs
```diff
@@ -192,6 +192,38 @@ impl Guard {
             local.flush(self);
         }
     }
+
+    /// Unpins and then pins the current thread.
+    ///
+    /// Since `&mut self` is given, we can statically deduce that there is no [`Shared`] pointer
+    /// that is created before an invocation of `safepoint()` and then survives after the
+    /// invocation.
+    ///
```
ghost: Should probably also mention that the current thread will be repinned only if this is the only active guard.

ghost (approved):
LGTM after adding a comment.

ghost:
By the way, do you think your queue would be better served with HP instead of EBR for memory reclamation? The published paper says they used HP - probably for a good reason?

jeehoonkang:
I just rebased and added a comment.

I think HP is more friendly for wait-free constructions.  In almost every wait-free construction, a data structure (usually linked list) contains all the threads, and a thread has a pointer to another thread in the data structure.  "A points to B" means that A will help B in the next turn.  That help pointer is usually long-living: it is updated only if A succeeds an operation, and the pointer points to a valid object until an iteration of the data structure is over.  So If this pointer is managed by EBR, the global epoch will be very slowly advanced.

I'm trying to deal with long-living pointers with crossbeam-epoch by designating an independent collector that only deals with the help pointers.  It's fine to slowly advance its epoch, because there will not be so many help-pointers.  (WARNING: it's not finished, so I still don't know whether it works.)

I'm also interested in mixing HP and EBR as discussed in https://github.com/crossbeam-rs/rfcs/issues/8.  Maybe MSR's Snowflake is an interesting example of that direction: https://www.microsoft.com/en-us/research/publication/project-snowflake-non-blocking-safe-manual-memory-management-net/

Amanieu:
I actually implemented something similar: https://github.com/Amanieu/crossbeam-epoch/commit/7f535bdb57bbb5c8f5b99a3b48b07be1cfc1c533

src/internal.rs
```diff
@@ -283,14 +288,30 @@ impl Local {
...
+    pub fn repin(&self) {
+        let guard_count = self.guard_count.get();
+
+        // Update the local epoch only if there's only one guard.
+        if guard_count == 1 {
+            let epoch = self.epoch.load(Ordering::Relaxed);
+            let global_epoch = self.global().epoch.load(Ordering::Relaxed);
+
+            if epoch != global_epoch {
+                self.update_epoch(epoch, global_epoch, Ordering::Release);
```
Amanieu: A `SeqCst` fence isn't required here, a `Release` store should be enough to ensure correct ordering.

- We need to ensure any loads/stores from the previous epoch do not leak into the new one.
- However we don't care if any loads/store from the new epoch happen before the epoch counter is updated. At worse , other threads will see us in an older epoch and will delay GC slightly.

jeehoonkang: @Amanieu you're right.  Thanks for pointing out that!  Also it seems I forgot what I said a month ago: https://github.com/crossbeam-rs/rfcs/issues/18#issuecomment-336768214  😭

jeehoonkang:
I rebased and addressed @Amanieu's comment.  Thanks!

Amanieu:
I think the consensus from #45 is that this function should be renamed to `repin`.

jeehoonkang:
I rebased and changed the comment for `repin()`.

src/guard.rs
```diff
@@ -203,10 +240,6 @@ impl Guard {
     ///
     /// If this method is called from an [`unprotected`] guard, then the passed function is called
     /// directly without unpinning the thread.
-    ///
-    /// # Examples
-    ///
-    /// ```
```
Amanieu: You accidentally deleted part of a comment here.

PR #44, opened 2017-11-24
Fix issues in sync/list and rename CasError.previous to current

Vtec234:
This PR fixes two bugs in `sync/list.rs`, one of them being due to the wrong field of a tuple returned from `compare_and_set` being used [here](https://github.com/crossbeam-rs/crossbeam-epoch/blob/a25ef9dc2dc0239d062d921da7e040c2477b1795/src/sync/list.rs#L159) (unnamed fields are evil!) and the other [here](https://github.com/crossbeam-rs/crossbeam-epoch/blob/a25ef9dc2dc0239d062d921da7e040c2477b1795/src/sync/list.rs#L216). A few tests to catch these are added. Additionally, it renames `Container` to `IsElement` and makes it use references instead of pointers.

Then, I also renamed `CasError.previous` to `current`, the reasoning being that at the time of CAS failure this value is still in the pointer, so it's "currently" there.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -114,8 +114,9 @@ impl Global {
         for local in self.locals.iter(&guard) {
             match local {
                 Err(IterError::Stalled) => {
-                    // A concurrent thread stalled this iteration. Since it also tries to advance
-                    // the epoch, we leave the job to that thread.
+                    // A concurrent thread stalled this iteration. That thread might also try to
+                    // advance the epoch, in which case we leave the job to it. Otherwise, the
+                    // epoch will not be advanced.
                     return global_epoch;
                 }
                 Ok(local) => {
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -114,8 +114,8 @@ impl Global {
         for local in self.locals.iter(&guard) {
             match local {
                 Err(IterError::Stalled) => {
-                    // The iteration is stalled by another thread's iteration. Since that thread
-                    // also tries to advance the epoch, we leave the job to that thread.
+                    // A concurrent thread stalled this iteration. Since it also tries to advance
+                    // the epoch, we leave the job to that thread.
                     return global_epoch;
                 }
                 Ok(local) => {
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -31,13 +31,13 @@ use collector::Handle;
 use epoch::{AtomicEpoch, Epoch};
 use guard::{unprotected, Guard};
 use garbage::{Bag, Garbage};
-use sync::list::{List, Entry, IterError, Container};
+use sync::list::{List, Entry, IterError, IsElement};
 use sync::queue::Queue;
 
 /// The global data for a garbage collector.
 pub struct Global {
     /// The intrusive linked list of `Local`s.
-    locals: List<Local, LocalContainer>,
+    locals: List<Local>,
 
     /// The global queue of bags of deferred functions.
     queue: Queue<(Epoch, Bag)>,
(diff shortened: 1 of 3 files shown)


src/sync/list.rs
```diff
@@ -213,7 +239,7 @@ impl<'g, T: 'g, C: Container<T>> Iterator for Iter<'g, T, C> {
                 let succ = succ.with_tag(0);
 
                 match self.pred.compare_and_set_weak(
-                    self.curr,
+                    self.curr.with_tag(0),
```
jeehoonkang: What if we immediately return `IterError::Stalled` when `self.curr.tag() != 0`?  Because nonzero `self.curr.tag()` means the node containing `self.pred` is deleted, and the iteration is stalled.

src/sync/list.rs
```diff
@@ -213,7 +239,7 @@ impl<'g, T: 'g, C: Container<T>> Iterator for Iter<'g, T, C> {
                 let succ = succ.with_tag(0);
 
                 match self.pred.compare_and_set_weak(
```
jeehoonkang: How about using strong `compare_and_set` here?  We want to advance the global epoch as much as possible.

jeehoonkang (commented):
First of all, thank you for catching this nasty bug and improving the code quality a lot!

Besides the comments I left so far, I have another one on documentation.  Currently a comment on `enum IterError::Stalled` says:

```rust
    /// Iterator was stalled by another iterator. Internally, the thread lost a race in deleting a
    /// node by a concurrent thread.
    Stalled,
```

However, as you pointed out, actually the iteration may also be stalled when a node (namely `self.pred`) is removed from the list.  How about saying that the iteration is stalled for an unknown reason, and you should reiterate over the list from the beginning?

Also, `Global::try_advance()` says:

```rust
                Err(IterError::Stalled) => {
                    // The iteration is stalled by another thread's iteration. Since that thread
                    // also tries to advance the epoch, we leave the job to that thread.
                    return global_epoch;
```

But actually it is stalled for an unknown reason, and it just gave up to advance the global epoch.

src/sync/list.rs
```diff
@@ -223,14 +249,14 @@ impl<'g, T: 'g, C: Container<T>> Iterator for Iter<'g, T, C> {
...
-                            self.guard.defer(move || C::finalize(p.as_raw()));
+                            self.guard.defer(move || C::finalize(p.deref()));
                         }
                         self.curr = succ;
                     }
                     Err(err) => {
                         // We lost the race to delete the entry by a concurrent iterator. Set
                         // `self.curr` to the updated pointer, and report that we are stalled.
-                        self.curr = err.new;
+                        self.curr = err.current;
```
ghost: @jeehoonkang Shouldn't we start again from the head here?

Vtec234: You're right, we should! **EDIT:** Well, I think this behaviour is also correct (won't break the list), but it does not meet the specification which says that we do restart.

src/sync/list.rs
```diff
@@ -212,37 +241,50 @@ impl<'g, T: 'g, C: Container<T>> Iterator for Iter<'g, T, C> {
...
+                    // A concurrent thread deleted the predecessor node - we are stalled
+                    // and have to restart from `head`.
+                    self.pred = self.head;
+                    self.curr = self.head.load(Acquire, self.guard);
+
+                    return Some(Err(IterError::Stalled));
+                }
+
+                match self.pred.compare_and_set(
+                    self.curr.with_tag(0),
```
ghost: `.with_tag(0)` is not necessary anymore, but you can leave it.

ghost (approved):

src/internal.rs
```diff
@@ -114,8 +114,8 @@ impl Global {
         for local in self.locals.iter(&guard) {
             match local {
                 Err(IterError::Stalled) => {
-                    // The iteration is stalled by another thread's iteration. Since that thread
-                    // also tries to advance the epoch, we leave the job to that thread.
+                    // A concurrent thread stalled this iteration. Since it also tries to advance
```
jeehoonkang: The concurrent thread may not try to advance the global epoch; it may just delete a node.

src/sync/list.rs
```diff
@@ -212,37 +241,50 @@ impl<'g, T: 'g, C: Container<T>> Iterator for Iter<'g, T, C> {
                 // This entry was removed. Try unlinking it from the list.
                 let succ = succ.with_tag(0);
 
-                match self.pred.compare_and_set_weak(
+                if self.curr.tag() != 0 {
```
jeehoonkang: After changing the behavior of the `Err` branch below (restarting from the beginning), it seems `self.cur.tag()` should always be zero.  If so, we can remove this branch.

src/internal.rs
Vtec234: Since we can't tell whether the concurrent thread is just deleting a node or actually advancing the epoch, should the code stay the same and the comment just be updated to document this?

src/sync/list.rs
Vtec234: It seems you are right! Updated.

jeehoonkang (approved):

PR #45, opened 2017-11-25
Add the ability to temporarily unpin a Guard

Amanieu:
This is similar to #43, but allows a function to be executed with the guard unpinned. I need this functionality for my use case where a thread will stay pinned most of the time, except for some potentially long-running system calls.

### DIFF
--- a/src/guard.rs
+++ b/src/guard.rs
@@ -198,7 +198,7 @@ impl Drop for Guard {
     #[inline]
     fn drop(&mut self) {
         if let Some(local) = unsafe { self.local.as_ref() } {
-            Local::unpin(local);
+            local.unpin();
         }
     }
 }
--- a/src/guard.rs
+++ b/src/guard.rs
@@ -197,8 +197,9 @@ impl Guard {
     /// Temporarily unpins the thread, executes the given function and then re-pins the thread.
     ///
     /// This method is useful when you need to perform a long-running operation (e.g. sleeping)
-    /// and don't need to maintain any guard-based reference across the call. The thread will only
-    /// be unpinned if this is the only active guard for the current thread.
+    /// and don't need to maintain any guard-based reference across the call (the latter is enforced
+    /// by `&mut self`). The thread will only be unpinned if this is the only active guard for the
+    /// current thread.
     ///
     /// If this method is called from an [`unprotected`] guard, then the passed function is called
     /// directly without unpinning the thread.
@@ -217,7 +218,7 @@ impl Guard {
     ///     let p = a.load(SeqCst, &guard);
     ///     assert_eq!(unsafe { p.as_ref() }, Some(&777));
     /// }
-    /// guard.unpin_for(|| thread::sleep(Duration::from_millis(50)));
+    /// guard.repin_after(|| thread::sleep(Duration::from_millis(50)));
     /// {
     ///     let p = a.load(SeqCst, &guard);
     ///     assert_eq!(unsafe { p.as_ref() }, Some(&777));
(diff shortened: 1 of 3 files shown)


ghost:
Nice -- this seems more powerful than `.safepoint()`, although re-pinning this way is a bit slower.

A few weeks ago, when I brought up the idea of safepoints we still had `epoch::pin(|scope| ...)` instead of `let guard = epoch::pin()`. The thing about scopes is that you cannot drop the `scope` passed to the closure (its type is `&Scope`), but it's possible to manipulate guards in a similar manner.

For example, with guards you can do the following:

```rust
let mut guard = Some(epoch::pin());

for _ in 0..100 {
    {
        let guard = guard.as_ref().unwrap();
        let p = my_atomic.load(SeqCst, &guard);
    }

    // Drop the guard.
    guard = None;

    // Perform a long-running system call, computation, or whatever...

    // Re-pin.
    guard = Some(epoch::pin());
}
```

That code effectively does the same thing as `.unpin_for()` so I wonder if we need that method at all. Is there any advantage of `.unpin_for()` that I'm missing?

One advantage of `.safepoint()` is that re-pinning is a bit faster than `guard = None; guard = Some(epoch::pin())`, although we haven't measured how much yet.

Now I'm also wondering whether we even need `.safepoint()`... :)

Amanieu:
I actually considered the `Option` approach, but having to unwrap the all the time becomes messy, especially when you know (but the compiler doesn't) that it contains `Some(Guard)`.

Regarding `safepoint`, I think we should have **both** `unpin_for` and `safepoint` since they serve different purposes and have very different performance characteristics (`safepoint` is just a single `store(Release)`).

URCU's [QSBR API](https://github.com/urcu/userspace-rcu/blob/d4e640c06c2c999d5fc8f8375bc2f61da06c9cda/README.md#usage-of-liburcu-qsbr) has a similar distinction between these two use cases:

- `safepoint` is roughly equivalent to `rcu_quiescent_state`.
- `unpin_for` is roughly equivalent to `rcu_thread_online`/`rcu_thread_offline`.

By the way I am not completely sold on the `unpin_for` name and don't mind changing it for something better.

ghost:
That's fair enough - let's add this method.

Regarding naming, my only complaint is that `safepoint` and `unpin_for` have totally different names, while they're very similar methods.

Perhaps we should call them `safepoint` and `safepoint_for`?
Or maybe `quiescent` and `quiescent_for`?

Amanieu:
I don't have a particular preference for the name, but I agree with you that the two functions should have similar names.

src/guard.rs
```diff
@@ -192,13 +193,70 @@ impl Guard {
...
+        F: FnOnce() -> R,
+    {
+        if let Some(local) = unsafe { self.local.as_ref() } {
+            // We need to acquire a handle here to ensure the Local doesn't
+            // disappear from under us.
+            local.acquire_handle();
+            local.unpin();
+        }
+
+        // Ensure the Guard is re-pinned even if the function panics
```
jeehoonkang, line 240: We used to use `scopeguard` crate for this purpose: https://crates.io/crates/scopeguard  I prefer to use it because it's easier to read.  What do you think?  cc @stjepang

jeehoonkang (commented):
I prefer `quiescient/quiescient_for` to `safepoint/safepoint_for`, because `safepoint_for` seems ambiguous.

But `quiescient` is a difficult word (though widely used in RCU), I want to come up with an eaiser name.  Maybe `repin/repin_after`?

src/guard.rs
Amanieu: I changed it to use `defer!` from scopeguard.

ghost (approved):
Let's rename to `repin_after` and then we can merge.

Amanieu:
Rebased.

jeehoonkang (approved):

PR #46, opened 2017-11-25
Add no_std support

Amanieu:
This PR enables no_std support for crossbeam-epoch. It depends on crossbeam-rs/crossbeam-utils#3.

The only API-level change is that a default collector is only provided if the "std" feature is enabled (it is enabled by default).

Note that no_std is only properly supported if the "std" feature is disabled **and** the "nightly" feature is enabled. This is because this crate uses `Box` and `Arc` which depend on liballoc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,14 +12,13 @@ keywords = ["lock-free", "rcu", "atomic", "garbage"]
 categories = ["concurrency", "memory-management"]
 
 [features]
-nightly = []
+default = ["std"]
+std = ["lazy_static"]
+nightly = ["arrayvec/use_union"]
 strict_gc = []
 
 [dependencies]
-arrayvec = "0.4"
-crossbeam-utils = "0.1"
-lazy_static = "0.2"
-memoffset = "0.1.0"
-
-[dev-dependencies]
-rand = "0.3"
+arrayvec = { version = "0.4", default-features = false }
+crossbeam-utils = { version = "0.1", default-features = false }
+lazy_static = { version = "0.2", optional = true }
+memoffset = { version = "0.1", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ keywords = ["lock-free", "rcu", "atomic", "garbage"]
 categories = ["concurrency", "memory-management"]
 
 [features]
-default = ["std"]
-std = ["lazy_static"]
+default = ["use_std"]
+use_std = ["lazy_static"]
 nightly = ["arrayvec/use_union"]
 strict_gc = []
 
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -13,10 +13,7 @@
 /// handle.pin().flush();
 /// ```
 
-#[cfg(feature = "nightly")]
 use alloc::arc::Arc;
-#[cfg(not(feature = "nightly"))]
-use alloc::sync::Arc;
 
 use internal::{Global, Local};
 use guard::Guard;
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -13,7 +13,10 @@
 /// handle.pin().flush();
 /// ```
 
-use std::sync::Arc;
+#[cfg(feature = "nightly")]
+use alloc::arc::Arc;
+#[cfg(not(feature = "nightly"))]
+use alloc::sync::Arc;
 
 use internal::{Global, Local};
 use guard::Guard;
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -12,14 +12,13 @@ keywords = ["lock-free", "rcu", "atomic", "garbage"]
 categories = ["concurrency", "memory-management"]
 
 [features]
-nightly = []
+default = ["std"]
+std = ["lazy_static"]
```
ghost: Really minor nit: Is there a widely accepted convention for what this feature should be named? It seems the following crates call it `use_std` rather than just `std`:

* [libc](https://github.com/rust-lang/libc/blob/master/Cargo.toml)
* [futures](https://github.com/alexcrichton/futures-rs/blob/master/Cargo.toml)
* [either](https://github.com/bluss/either/blob/master/Cargo.toml#L23)

On the other hand, [failure](https://github.com/withoutboats/failure/blob/master/Cargo.toml) does call it `std`...

ghost:
Looks good to me.

I've never worked much with `no_std`, so pardon my ignorance for asking:

1. Would it make sense for `crossbeam-epoch` to allow `no_std` support on stable Rust as well?
2. What's the stabilization plan for the liballoc? Is liballoc going to be forever nightly-only? I'm aware that things like `Alloc` trait and `Heap` are going to be stabilized soon, but are they going to be simply reexported into libstd, or will liballoc be available on stable Rust as well?

Cargo.toml
Amanieu: I don't think there's really a universal convention. It probably doesn't matter since nobody will use that feature name directly: anyone wanting `no_std` will just use `default-features = false`.

Amanieu:
1. I don't think `no_std` is possible with only stable rust since the code relies on `Arc` and `Box`, which are only available through libstd on stable.

2. My understanding is that liballoc will eventually be stabilized, but I'm not sure if there is a timeline for this.

In any case, most people currently working with no_std environments use nightly compilers, so I don't think it should be too much of an issue.

ghost (approved):

src/collector.rs
```diff
@@ -13,7 +13,10 @@
 /// handle.pin().flush();
 /// ```
 
-use std::sync::Arc;
+#[cfg(feature = "nightly")]
```
jeehoonkang: Maybe the `cfg_if` crate would be useful for implementing this branch.  @stjepang used it in https://github.com/crossbeam-rs/crossbeam-epoch/pull/39/files#diff-7f3d2612bf5d28e2e8a5055f7ecfdc8fR32

src/internal.rs
```diff
@@ -16,13 +16,17 @@
...
-use std::sync::atomic;
-use std::sync::atomic::Ordering::{Acquire, Relaxed, Release, SeqCst};
+use core::cell::{Cell, UnsafeCell};
+use core::mem::{self, ManuallyDrop};
+use core::num::Wrapping;
+use core::ptr;
+use core::sync::atomic;
+use core::sync::atomic::Ordering::{Acquire, Relaxed, Release, SeqCst};
+use alloc::boxed::Box;
+#[cfg(feature = "nightly")]
```
jeehoonkang: As above, `cfg_if` would be helpful.

src/lib.rs
```diff
@@ -55,16 +55,32 @@
...
+#[macro_use]
+extern crate std;
+#[cfg(test)]
+extern crate core;
+
+// Use liballoc on nightly to avoid a dependency on libstd
+#[cfg(feature = "nightly")]
+extern crate alloc;
+#[cfg(not(feature = "nightly"))]
+extern crate std as alloc;
```
jeehoonkang: Even if we rename `std` into `alloc` here, there's another branch on `cfg(feature = "nightly")` in importing `Arc`.  If so, how about:

- just `extern crate std` here;
- import `Box` and `Arc` in `lib.rs` depending on `fg(feature = "nightly")`;
- use `epoch::Box` and `epoch::Arc` elsewhere?

jeehoonkang (commented):
This looks great!  I didn't notice that we can reduce `std`-dependent code as far as this PR.  I just left minor comments.  Thanks.

Cargo.toml
jeehoonkang: I noticed that we control whether to use `std` or not with `default-features = false`.  To my eyes `std` and `default` seems quite unrelated..  Is it an accepted convention?

jeehoonkang (approved):
I left another comment on `default-features`, but except for that I'm happy with this PR.  Thanks!

Cargo.toml
Amanieu: Yes, this seems to be a widely used convention. I'm not too happy about it, but there isn't much else that we can do with Cargo.

Also, after looking at a few crates, it seems that `use_std` is actually the standard feature name for rust-lang-nursery crates, so I'll just use that.

jeehoonkang:
I'm happy to merge this now.  @stjepang please have another look and merge!

PR #48, opened 2017-11-26
Update crossbeam-util version to 0.2 for no_std support

Amanieu:
Missed this in #46.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["concurrency", "memory-management"]
 
 [features]
 default = ["use_std"]
-use_std = ["lazy_static"]
+use_std = ["lazy_static", "crossbeam-utils/use_std"]
 nightly = ["arrayvec/use_union"]
 strict_gc = []
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ strict_gc = []
 
 [dependencies]
 arrayvec = { version = "0.4", default-features = false }
-crossbeam-utils = { version = "0.1", default-features = false }
+crossbeam-utils = { version = "0.2", default-features = false }
 lazy_static = { version = "0.2", optional = true }
 memoffset = { version = "0.1", default-features = false }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ strict_gc = []
 
 [dependencies]
 arrayvec = { version = "0.4", default-features = false }
-crossbeam-utils = { version = "0.1", default-features = false }
+crossbeam-utils = { version = "0.2", default-features = false }
```
jeehoonkang, line 22: I think we need `crossbeam-utils` 0.2.1, right?

Amanieu: I normally only put the major version in Cargo.toml. Cargo will automatically use the latest compatible version (0.2.1, 0.2.2, etc).

ghost (approved):

jeehoonkang:
Not sure why, but Travis complains about this PR: https://travis-ci.org/crossbeam-rs/crossbeam-epoch/jobs/307518318#L556

Amanieu:
Oh, it seems that the tests rely on scoped thread support. I can work around that...

ghost:
@jeehoonkang Tests pass on Travis now. Feel free to merge if you're happy with the PR.

Cargo.toml
jeehoonkang: I'm aware that specifying version = "0.2" works and it is an accepted convention to ignore the patch version in specifying dependency.  However, in my understanding, it's an undocumented behavior of `cargo`, and pedantically speaking, a future version of `cargo` may change its dependency resolution algorithm and fetches "0.2.0".

For this reason, I prefer to specify version = "0.2.1".

But I'm also aware that this matter is really minor: it's highly likely that specifying version = "0.2" will work correctly forever.  Since we're in front of a big 0.1.0 release, it's more beneficial to merge this PR now.  I'm approving and merging this PR.

jeehoonkang (approved):

PR #51, opened 2017-11-27
Fix a bug in calculating the epoch distance

### LINKED ISSUE #50: Using `Epoch::distance()` may give us a wrong result?
Suppose that `usize` is 32bits.  Then the implementation of `Epoch::decompose()` implies the global epoch wraps around in 31bits: https://github.com/crossbeam-rs/crossbeam-epoch/blob/master/src/epoch.rs#L71

Then to my eyes, `Epoch::distance()` may give us a wrong result: https://github.com/crossbeam-rs/crossbeam-epoch/blob/master/src/epoch.rs#L39 .  Suppose `e1` is the maximum number that is represented in 31bits (i.e. `(1 << 31) - 1`), and `e2` is the next epoch, which is represented in `0`.  Then `distance()` gives us `(1 << 31) - 1`, while a desired result is `1`.

This will probably happen in a long-running process in 32bit machines.

I'd like to propose to change `Epoch::distance()` as follows:

```rust
    fn decompose(&self) -> (usize, bool) {
        (self.data & !1, (self.data & 1) == 1)
    }
```

Then the stride of the epoch advancement becomes `2`, and we should adjust the implementation a little bit.

jeehoonkang:
Closes #50

### DIFF
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -169,7 +169,7 @@ mod tests {
                             collector.global.collect(guard);
                             let after = collector.global.epoch.load(Ordering::Relaxed);
 
-                            assert!(after.distance(before) <= 2);
+                            assert!(after.wrapping_sub(before) <= 4);
                         }
                     })
                 })
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -169,7 +169,7 @@ mod tests {
                             collector.global.collect(guard);
                             let after = collector.global.epoch.load(Ordering::Relaxed);
 
-                            assert!(after.wrapping_sub(before) <= 4);
+                            assert!(after.wrapping_sub(before) <= 2);
                         }
                     })
                 })
--- a/src/epoch.rs
+++ b/src/epoch.rs
@@ -7,7 +7,6 @@
 //! If an object became garbage in some epoch, then we can be sure that after two advancements no
 //! participant will hold a reference to it. That is the crux of safe memory reclamation.
 
-use core::cmp;
 use core::sync::atomic::{AtomicUsize, Ordering};
 
 /// An epoch that can be marked as pinned or unpinned.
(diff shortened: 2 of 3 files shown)


src/epoch.rs
```diff
@@ -27,49 +26,38 @@ impl Epoch {
...
+    pub fn unpinned(self) -> Epoch {
         Epoch { data: self.data & !1 }
     }
 
     /// Returns the successor epoch.
     ///
     /// The returned epoch will be marked as pinned only if the previous one was as well.
     #[inline]
-    pub fn successor(&self) -> Epoch {
+    pub fn successor(self) -> Epoch {
```
ghost, line 62: Good riddance of references, I like this.

ghost (changes requested):
I find the behavior of `wrapping_sub` confusing - and the documentation of this method is not particularly illuminating either. :)

How would you explain what `a.wrapping_sub(b)` does exactly? The answer is basically: "It transmutes `Epoch`s `a` and `b` into `usize`s and then subtracts them". That is an operation with very subtle behavior, which makes it easy to misuse.

For example, the method will return 1 for 

```rust
let a = Epoch::starting().pinned();
let b = Epoch::starting();
a.wrapping_sub(b)
```

But what does the result of 1 mean?

The old `distance` method was pretty clear: it computes the distance between two epochs (regardless of direction), that's it. No subtle tricks with strides of two - adjacent epochs differ by one. And the *pinned* flag is completely ignored.

Also, keep in mind that consecutive epochs don't *actually* come in strides of 2. They should differ by 1. The fact that we use the least significant bit to represent the *pinned* flag is just an arbitrary choice - it's an implementation detail. It could have been the most significant bit just as well! And this implementation detail shouldn't leak outside the interface of the `Epoch` struct.

It's especially confusing when `a.wrapping_sub(b) >= 4` actually means "if a is at least 2 epochs ahead of b".

If you don't like `distance` because it's bidirectional (I don't like it much either), then I'd suggest changing `wrapping_sub` as follows.

```rust
/// Returns the number of epochs `self` is ahead of `rhs`.
/// 
/// Internally, epochs are represented as numbers in the range `isize::MIN / 2 ..= isize::MAX / 2`,
/// so the returned distance will be in the same interval.
fn wrapping_sub(self, rhs: Self) -> isize {
    (self.data & !1).wrapping_sub(rhs.data & !1) as isize >> 1
}
```

Also, thinking of epochs as `isize`s instead of `usize`s makes it easier to express things like:

```rust
a.wrapping_sub(b) >= 2 // `a` is at least two epochs ahead of `b`
a.wrapping_sub(b) <= -2 // `a` is at least two epochs behind `b`
```

jeehoonkang:
@stjepang I changed `Epoch::wrapping_sub` as you suggested.  I also thought of the problem you mentioned, but I implemented `wrapping_sub` as I first did because I worried of the additional cost of bit-and & shift instructions.  However, the cost should be negligible..

ghost (approved):

jeehoonkang:
The Travis CI failure is due to #52

Vtec234 (commented):
Thanks!

Vtec234 (commented):
Thanks!

Vtec234 (approved):
Thanks! (not sure what happened there)

src/collector.rs
```diff
@@ -169,7 +169,7 @@ mod tests {
                             collector.global.collect(guard);
                             let after = collector.global.epoch.load(Ordering::Relaxed);
 
-                            assert!(after.distance(before) <= 2);
+                            assert!(after.wrapping_sub(before) <= 2);
```
Amanieu, line 172: Should this assert be changed from 2 to 1?
EDIT: nevermind, I misread.

src/epoch.rs
```diff
@@ -27,49 +26,39 @@ impl Epoch {
...
-    pub fn distance(&self, other: Self) -> usize {
-        let (e1, _) = self.decompose();
-        let (e2, _) = other.decompose();
-        cmp::min(e1.wrapping_sub(e2), e2.wrapping_sub(e1))
+    /// Returns the number of epochs `self` is ahead of `rhs`.
+    /// 
+    /// Internally, epochs are represented as numbers in the range `(isize::MIN / 2) .. (isize::MAX
+    /// / 2)`, so the returned distance will be in the same interval.
+    pub fn wrapping_sub(self, rhs: Self) -> isize {
+        (self.data & !1).wrapping_sub(rhs.data & !1) as isize >> 1
```
Vtec234: I realized you probably don't need the first (`& !1`) because it's shifted away. The second one has to stay.

jeehoonkang: Right.  I expected LLVM InstCombine will optimize it out, but I could confirm that it didn't.

I pushed a patch that removes the first `& !1`.  But I'm worrying about the readability: even though I left a comment, it may be strange to someone's eyes.  Any thoughts?  cc @stjepang

Issue #52, opened 2017-11-28
`cargo build` fails in Rust Nightly 1.24.0 (2017-11-27)


jeehoonkang:
```
$ rustc --version
rustc 1.24.0-nightly (560a5da9f 2017-11-27)

$ cargo build
   Compiling crossbeam-epoch v0.1.0 (file:///Users/jeehoon.kang/Works/crossbeam-epoch)
error[E0119]: conflicting implementations of trait `core::convert::Into<std::boxed::Box<_>>` for type `atomic::Owned<_>`:
   --> src/atomic.rs:732:1
    |
732 | / impl<T> Into<Box<T>> for Owned<T> {
733 | |     /// Converts the owned pointer into a `Box`.
734 | |     ///
735 | |     /// # Examples
...   |
748 | |     }
749 | | }
    | |_^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T, U> core::convert::Into<U> for T
              where U: core::convert::From<T>;

error: aborting due to previous error

error: Could not compile `crossbeam-epoch`.

To learn more, run the command again with --verbose.
```

ghost:
Looks like https://github.com/rust-lang/rust/issues/46328

ghost:
Unfortunately, this error will persist. See [this comment](https://github.com/rust-lang/rust/issues/46328#issuecomment-347557526) for an explanation.

Let's bring back the explicit `into_box` method which we used to have a few weeks ago.

By the way, if you're worried that `into_box` might be unidiomatic, note that there are already several similar methods in the standard library:
https://doc.rust-lang.org/alloc/raw_vec/struct.RawVec.html#method.into_box
https://doc.rust-lang.org/collections/vec/struct.Vec.html#method.into_boxed_slice
https://doc.rust-lang.org/std/string/struct.String.html#method.into_boxed_str

PR #53, opened 2017-11-28
Fix an ordering bug in Global::push_bag()

jeehoonkang:
In an RFC (https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-07-23-relaxed-memory.md), we explained why Crossbeam is correct in relaxed memory.  However, the pseudocode in the RFC and the implementation differs: in `unlink()` in RFC, a `SeqCst` fence is executed and then the global epoch is read.  On the other hand, in `push_bag()` in the implementation, the global epoch is read and then a `SeqCst` fence is executed.

Unfortunately, the current implementation is incorrect.  For example, consider the following scenario:

- Thread A is pinned at epoch 10, and deferring the deallocation of an object O.
- In `Global::push_bag()`, A can read the global epoch 10, and then sleeps for a long time.
- In the mean time, the global epoch advanced to 11.
- Thread B is pinned with local epoch 11, and got a reference to O.
- Thread A woke up and executes the `SeqCst` fence, and then it is unpinned.
- The global epoch advances to 12.
- Thread C actually deallocates O, while B has a reference to it...

This PR fixes it by reordering `fence(SeqCst)` and a load from the global epoch in `Global::push_bag()`.

Arguably, the performance remains almost the same after this PR.  This is an one-time measurement of `cargo bench` (control = master, variable = this PR):

```
 name                       control ns/iter  variable ns/iter  diff ns/iter  diff %  speedup
 multi_alloc_defer_free     3,581,149        3,608,989               27,840   0.78%   x 0.99
 multi_defer                1,640,773        1,630,437              -10,336  -0.63%   x 1.01
 multi_flush                7,538,490        7,519,137              -19,353  -0.26%   x 1.00
 multi_pin                  4,725,144        4,674,670              -50,474  -1.07%   x 1.01
 single_alloc_defer_free    49               53                           4   8.16%   x 0.92
 single_default_handle_pin  9                9                            0   0.00%   x 1.00
 single_defer               27               30                           3  11.11%   x 0.90
 single_flush               214              206                         -8  -3.74%   x 1.04
 single_pin                 9                9                            0   0.00%   x 1.00
```

This PR is blocked by #52: here I just commented out `impl<T> Into<Box<T>> for Owned<T>`, but it should be properly dealt with in another PR.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -63,11 +63,11 @@ impl Global {
 
     /// Pushes the bag into the global queue and replaces the bag with a new empty bag.
     pub fn push_bag(&self, bag: &mut Bag, guard: &Guard) {
-        let epoch = self.epoch.load(Ordering::Relaxed);
         let bag = mem::replace(bag, Bag::new());
 
         atomic::fence(Ordering::SeqCst);
 
+        let epoch = self.epoch.load(Ordering::Relaxed);
         self.queue.push((epoch, bag), guard);
     }
 


ghost (approved):
Good find! :)

martinhath:
I must say I'm a little confused by this (although it's probably correct :).
Presumably the problem is that A read an old epoch in `unlink`, and B has read the new epoch, and manages to obtain a reference to `O`, before As unlinking is visible to B?

Does the fence in `unlink` guarantee that after executing it, is it not possible for any thread to obtain a reference to `O`, such that we know if the epoch is incremented to 11, and thread B has observed this, the read by A will also read 11?

I must admit the semantics of `fence` is really confusing to me; the docs talks about synchronization between operations happening before and after, but here it seems all operations are after the fence.

jeehoonkang:
> I must admit the semantics of `fence` is really confusing to me; the docs talks about synchronization between operations happening before and after, but here it seems all operations are after the fence.

The semantics of `fence(SeqCst)` has long been unclear, and only recently is it clarified.
 See: http://sf.snu.ac.kr/promise-concurrency/ and http://plv.mpi-sws.org/scfix/  (disclosure: I'm one of the authors of both papers).


> Does the fence in `unlink` guarantee that after executing it, is it not possible for any thread to obtain a reference to `O`, such that we know if the epoch is incremented to 11, and thread B has observed this, the read by A will also read 11?

Assuming the semantics presented in those papers are correct, the semantics of `fence(SeqCst)` is weaker than you expect.  Even if A executes `fence(SeqCst)`, it doesn't necessarily mean all the threads should not be able to read `O`.  Only if another thread `B` executes `fence(SeqCst)` after A does, `B` cannot read the unlinked object.

Basically that's why we need to wait for 2 epoch advances.  The RFC I mentioned above (https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-07-23-relaxed-memory.md) explains in fuller details of its correctness.

martinhath:
> Even if A executes fence(SeqCst), it doesn't necessarily mean all the threads should not be able to read O. Only if another thread B executes fence(SeqCst) after A does, B cannot read the unlinked object.

I see. So in this PR, thread B is the one in `push_bag`, such that it must `fence` before reading, which then guarantees that it observes the epoch increment, which is also followed by a `fence`?

jeehoonkang:
@martinhath I'm sorry for the delay.

I'm sorry again that I couldn't fully understand your comment..  I presumed A is the one in `push_bag` (or `unlinked`): A's action (removal of `O` from the memory) should be visible when the epoch advanced enough in `try_advance()`.


Instead of directly answering your question, I'd like to summarize what's going on in the proof in https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-07-23-relaxed-memory.md .  In essence, the reasoning depends on the following two properties of SC fences:

- All the executions of SC fences are linearly ordered.
- If an execution A of SC fence is before another execution B in the linear order, A's knowledge before its fence should be transferred to B after its fence.

So the proof first (1) somehow deduces that `unlinked()`'s SC fence should be executed before `try_advance()`'s SC fence; so (2) `O` should be invisible after `try_advance()` successfully advanced the epoch enough.

I hope it helps, at least a little bit.  Please feel free to ask any questions.

martinhath:
Okay, I think I understand now. Thanks for your help! 😄

PR #54, opened 2017-11-28
Fix a bug in Local::finalize

ghost:
1. Fix a bug in  `Local::finalize` where we had `global.push_bag(&mut *self.bag.get(), &unprotected());`, which is completely wrong. Pushing a bag into the queue without pinning will cause use-after-free.

2. Change the sanitization test a little bit so that use-after-frees appear more often.

3. Switch from thread sanitizer to address sanitizer. Unfortunately, our code is full of fences, which thread sanitizer simply doesn't understand. I've noticed several false positives coming from thread sanitizer. The good news is that address sanitizer can still catch some errors due to bad memory orderings, incorrect epoch advancement, incorrect use of unprotected guards, etc.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -353,7 +353,8 @@ impl Local {
         // doesn't call `finalize` again.
         self.handle_count.set(1);
         unsafe {
-            // Pin and move the local bag into the global queue.
+            // Pin and move the local bag into the global queue. It's important that `push_bag`
+            // doesn't defer destruction on any new garbage.
             let guard = &self.pin();
             self.global().push_bag(&mut *self.bag.get(), guard);
         }
@@ -361,7 +362,9 @@ impl Local {
         self.handle_count.set(0);
 
         unsafe {
-            // Take the reference to the `Global` out of this `Local`.
+            // Take the reference to the `Global` out of this `Local`. Since we're not protected
+            // by a guard at this time, it's crucial that the reference is read before marking the
+            // `Local` as deleted.
             let global: Arc<Global> = ptr::read(&**self.global.get());
 
             // Mark this node in the linked list as deleted.
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -88,7 +88,13 @@ impl Global {
             global_epoch.distance(item.0) > 1
         };
 
-        for _ in 0..Self::COLLECT_STEPS {
+        let steps = if cfg!(feature = "sanitize") {
+            usize::max_value()
+        } else {
+            Self::COLLECT_STEPS
+        };
+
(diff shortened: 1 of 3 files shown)


src/internal.rs
```diff
@@ -349,13 +349,21 @@ impl Local {
         debug_assert_eq!(self.guard_count.get(), 0);
         debug_assert_eq!(self.handle_count.get(), 0);
 
+        // Temporarily increment handle count. This is required so that the following call to `pin`
+        // doesn't call `finalize` again.
+        self.handle_count.set(1);
+        unsafe {
+            // Pin and move the local bag into the global queue.
+            let guard = &self.pin();
+            self.global().push_bag(&mut *self.bag.get(), guard);
```
jeehoonkang: Aha, now I can see what's wrong with the current implementation.  The problem is `push_bag()` may access the global queue's internal pointers without protection.  This patch adds that required protection.

I'd like to leave a comment that `push_bag()` will not add a garbage in the local bag so that `self.bag` is really empty after this call.

src/internal.rs
```diff
@@ -349,13 +349,21 @@ impl Local {
...
+        self.handle_count.set(0);
+
         unsafe {
             // Take the reference to the `Global` out of this `Local`.
             let global: Arc<Global> = ptr::read(&**self.global.get());
 
-            // Move the local bag into the global queue.
-            global.push_bag(&mut *self.bag.get(), &unprotected());
-
             // Mark this node in the linked list as deleted.
```
jeehoonkang: How about hoisting this code before `let global: ...`?

jeehoonkang (commented):
Thanks you for finding this nasty bug!  I left two minor comments.

jeehoonkang:
Travis CI failure: https://travis-ci.org/crossbeam-rs/crossbeam-epoch/jobs/308576435#L1147  It says:

```
==10033==LeakSanitizer has encountered a fatal error.
==10033==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==10033==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
```

src/internal.rs
ghost: Added a comment.

src/internal.rs
ghost: You mean, moving the call to `delete` above `ptr::read`? That would create another use-after-free. 

Note that at this point we're not protected by a guard at all. If we mark the `Local` as deleted first, it might get garbage collected by other threads before we manage to read the `Arc<Global>`.

jeehoonkang (approved):

PR #60, opened 2018-01-03
Clarify comment and add Debug to CASError

Vtec234:
This is a small patch that:
- clarifies the behaviour of `compare_and_set`
- adds `Debug` to `CompareAndSetError`


ghost (approved):

Vtec234:
I would rather not merge this until we figure out why the build fails, even though it's most likely unrelated.

ghost:
Looks like sanitizer support is currently broken. I've filed an issue: https://github.com/rust-lang/rust/issues/47178

PR #61, opened 2018-01-13
Support Rust 1.13.0

### LINKED ISSUE #57: Supporting Rust 1.12 for Rayon
In order to use Crossbeam in Rayon, we need to support Rust 1.12.  More discussions are in: https://github.com/rayon-rs/rayon/pull/480 .

I hope it is the only remaining blocker for Rayon.

jeehoonkang:
Likely to clos.. #57.  We have a chance that Rayon's minimum required Rust version can be bumped to 1.14.0.

Funny that I cannot use the right word above and here: it will clos.. #57!

### DIFF
--- a/src/guard.rs
+++ b/src/guard.rs
@@ -73,10 +73,24 @@ use internal::Local;
 ///
 /// [`pin`]: fn.pin.html
 pub struct Guard {
-    pub(crate) local: *const Local,
+    local: *const Local,
 }
 
 impl Guard {
+    /// Creates a new guard from a pointer to `Local`.
+    ///
+    /// # Safety
+    ///
+    /// The `local` should be a valid pointer created by `Local::register()`.
+    pub unsafe fn new(local: *const Local) -> Guard {
+        Guard { local: local }
+    }
+
+    /// Accesses the internal pointer to `Local`.
+    pub unsafe fn get_local(&self) -> *const Local {
+        self.local
+    }
+
     /// Stores a function so that it can be executed at some point after all currently pinned
     /// threads get unpinned.
     ///
@@ -353,17 +367,24 @@ impl Clone for Guard {
 /// is very helpful.
 ///
 /// ```
-/// use crossbeam_epoch::{self as epoch, Atomic};
-/// use std::mem::ManuallyDrop;
+/// extern crate crossbeam_epoch as epoch;
+/// #[cfg(feature = "manually_drop")]
+/// mod nodrop {
+///     pub use std::mem::ManuallyDrop as NoDrop;
+/// }
+/// #[cfg(not(feature = "manually_drop"))]
+/// extern crate nodrop;
+///
+/// use std::ptr;
 /// use std::sync::atomic::Ordering::Relaxed;
 ///
 /// struct Stack<T> {
-///     head: Atomic<Node<T>>,
+///     head: epoch::Atomic<Node<T>>,
 /// }
 ///
 /// struct Node<T> {
-///     data: ManuallyDrop<T>,
-///     next: Atomic<Node<T>>,
+///     data: nodrop::NoDrop<T>,
+///     next: epoch::Atomic<Node<T>>,
 /// }
 ///
 /// impl<T> Drop for Stack<T> {
@@ -378,7 +399,8 @@ impl Clone for Guard {
 ///
 ///                 // Take ownership of the node, then drop its data and deallocate it.
 ///                 let mut o = node.into_owned();
-///                 ManuallyDrop::drop(&mut o.data);
+///                 let data = nodrop::NoDrop::into_inner(ptr::read(&o.data));
+///                 drop(data);
 ///                 drop(o);
 ///
 ///                 node = next;
--- a/src/guard.rs
+++ b/src/guard.rs
@@ -82,11 +82,13 @@ impl Guard {
     /// # Safety
     ///
     /// The `local` should be a valid pointer created by `Local::register()`.
+    #[doc(hidden)]
     pub unsafe fn new(local: *const Local) -> Guard {
         Guard { local: local }
     }
 
     /// Accesses the internal pointer to `Local`.
+    #[doc(hidden)]
     pub unsafe fn get_local(&self) -> *const Local {
         self.local
     }
@@ -367,27 +369,20 @@ impl Clone for Guard {
 /// is very helpful.
 ///
 /// ```
-/// extern crate crossbeam_epoch as epoch;
-/// #[cfg(feature = "manually_drop")]
-/// mod nodrop {
-///     pub use std::mem::ManuallyDrop as NoDrop;
-/// }
-/// #[cfg(not(feature = "manually_drop"))]
-/// extern crate nodrop;
-///
+/// use crossbeam_epoch::{self as epoch, Atomic};
 /// use std::ptr;
 /// use std::sync::atomic::Ordering::Relaxed;
 ///
-/// struct Stack<T> {
-///     head: epoch::Atomic<Node<T>>,
+/// struct Stack {
+///     head: epoch::Atomic<Node>,
 /// }
 ///
-/// struct Node<T> {
-///     data: nodrop::NoDrop<T>,
-///     next: epoch::Atomic<Node<T>>,
+/// struct Node {
+///     data: u32,
+///     next: epoch::Atomic<Node>,
 /// }
 ///
-/// impl<T> Drop for Stack<T> {
+/// impl Drop for Stack {
 ///     fn drop(&mut self) {
 ///         unsafe {
 ///             // Unprotected load.
@@ -397,11 +392,8 @@ impl Clone for Guard {
 ///                 // Unprotected load.
 ///                 let next = n.next.load(Relaxed, epoch::unprotected());
 ///
-///                 // Take ownership of the node, then drop its data and deallocate it.
-///                 let mut o = node.into_owned();
-///                 let data = nodrop::NoDrop::into_inner(ptr::read(&o.data));
-///                 drop(data);
-///                 drop(o);
+///                 // Take ownership of the node, then drop it.
+///                 drop(node.into_owned());
 ///
 ///                 node = next;
 ///             }
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ rust:
   - stable
   - beta
   - nightly
+  - 1.13.0
 
 addons:
   apt:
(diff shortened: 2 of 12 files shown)


jeehoonkang:
- The CI build for nightly fails for an unknown reason: https://travis-ci.org/crossbeam-rs/crossbeam-epoch/jobs/328551907#L1181

  Does anyone know the reason?


- The CI build for Rust 1.12.1 fails due to `arrayvec`: https://travis-ci.org/crossbeam-rs/crossbeam-epoch/jobs/328551908#L669

  I just asked the maintainer of `arrayvec` to support Rust 1.12.1: https://github.com/bluss/arrayvec/pull/90

jeehoonkang:
Now CI is happy with this branch, except for the issue @stjepang reported: https://github.com/rust-lang/rust/issues/47178

Vtec234 (approved):
If rayon developers are fine with that, great!

jeehoonkang:
In fact, this branch can be successfully compiled with Rust 1.13.

jeehoonkang:
It turns out that we can support Rust 1.13!  After merging https://github.com/bluss/arrayvec/pull/90 and this PR, I'll release a new version of `crossbeam-epoch`, make a companion PR in `crossbeam-deque`, and ask the Rayon maintainers to merge it.

jeehoonkang:
https://github.com/bluss/arrayvec/pull/90 is merged, and `arrayvec` is bumped to `0.4.7`.

src/guard.rs
```diff
@@ -73,10 +73,24 @@ use internal::Local;
...
+    ///
+    /// # Safety
+    ///
+    /// The `local` should be a valid pointer created by `Local::register()`.
+    pub unsafe fn new(local: *const Local) -> Guard {
+        Guard { local: local }
+    }
+
+    /// Accesses the internal pointer to `Local`.
+    pub unsafe fn get_local(&self) -> *const Local {
```
ghost: Shouldn't `new` and `get_local` be crate-local functions (`pub(crate)`)? But I guess the problem is that Rust 1.13 doesn't support `pub(restricted)`? Well, that sucks. :(

We should do one of three things:

1. Mark these functions with `#[doc(hidden)]` so that they don't appear in the documentation.
2. Merge `guard.rs` and `internal.rs` into a single module to resolve the visibility problems.

src/guard.rs
```diff
@@ -353,17 +367,24 @@ impl Clone for Guard {
...
 /// struct Stack<T> {
-///     head: Atomic<Node<T>>,
+///     head: epoch::Atomic<Node<T>>,
 /// }
 ///
 /// struct Node<T> {
-///     data: ManuallyDrop<T>,
-///     next: Atomic<Node<T>>,
+///     data: nodrop::NoDrop<T>,
+///     next: epoch::Atomic<Node<T>>,
```
ghost: Perhaps `nodrop::` and `epoch::` are a little bit too verbose? :)
`nodrop::NoDrop` feels redundant (it's just a name repeated twice), while `epoch::Atomic` differs from the common style (`epoch::pin()`, `epoch::default_handle()`, but just `Atomic<Node<T>>`).

src/guard.rs
```diff
@@ -353,17 +367,24 @@ impl Clone for Guard {
...
 /// ```
-/// use crossbeam_epoch::{self as epoch, Atomic};
-/// use std::mem::ManuallyDrop;
+/// extern crate crossbeam_epoch as epoch;
+/// #[cfg(feature = "manually_drop")]
+/// mod nodrop {
+///     pub use std::mem::ManuallyDrop as NoDrop;
+/// }
+/// #[cfg(not(feature = "manually_drop"))]
+/// extern crate nodrop;
```
ghost: I don't like `#[cfg(...)]` being visible inside documentation.

Can we write two variants of the doctest and gate them using a cfg attribute?

ghost (changes requested):
I feel conflicted about all this. Now we have basically downgraded to Rust from 2016. :)

My thought process is that we should do what we can in order to support Rayon by making some kind of compromise to accommodate its requirements. But it's also important that we don't downgrade the whole Crossbeam project to Rust 1.13 - only the parts which are necessary for Rayon.

There are many useful features that came after 1.13 we are already using or are very likely to start using. Here are some off the top of my head:

- `ManuallyDrop` - commonly used in data structures.
- `mem::needs_drop` - can optimize code by avoiding unnecessary calls to `Guard::defer`.
- extended `Cell` API - we don't use it right now, but it's a versatile building block.
- `pub(restricted)` - makes the module system a lot less annoying.
- `ThreadId` - can be handy for synchronization.
- `compiler_fence` - similar to an atomic fence, could be useful.

There are a few interesting features nearing stabilization:

- `repr(align)` - we want it for `CachePadded`.
- `LocalKey::try_with` - needed for `try_default_handle()`.
- `mem::unreachable` - can be useful for low-level optimization.

Features that may be stabilized this year:

- `integer_atomics` - obviously related to Crossbeam. :)
- unsafe pointer methods - makes working with pointer operations a lot easier.
- `alloc` - would be useful in my skip list and in `crossbeam-epoch`/`elfmalloc`.
- auto traits - a `NonCell` auto trait could optimize away unnecessary clones in a Bw-Tree.
- specialization - very useful for optimization.

There are also a few quality-of-life improvements that I'm looking forward to, although it's possible to live without them:

- new module system
- non-lexical lifetimes

Another thing to consider is that Rayon depends on `crossbeam-deque`, and the deque should really switch to a lightweight hazard pointer-based scheme instead of using `crossbeam-epoch`. We could even implement simple HPs inside the `crossbeam-deque` crate. I wonder if downgrading `crossbeam-epoch` is actually necessary in the end. :)

src/guard.rs
jeehoonkang: I added `use nodrop::NoDrop` and `use epoch::Atomic`, but `cargo test` complains for a strange reason:

```
---- guard::unprotected_1 stdout ----
	error[E0432]: unresolved import `epoch::Atomic`
  --> <anon>:10:5
   |
10 | use epoch::Atomic;
   |     ^^^^^^^^^^^^^ Did you mean `self::epoch`?

error[E0432]: unresolved import `nodrop::NoDrop`
  --> <anon>:11:5
   |
11 | use nodrop::NoDrop;
   |     ^^^^^^^^^^^^^^ Did you mean `self::nodrop`?

error: aborting due to previous error(s)

thread 'guard::unprotected_1' panicked at 'Box<Any>', ../src/librustc/session/mod.rs:191
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    guard::unprotected_1

test result: FAILED. 40 passed; 1 failed; 3 ignored; 0 measured

error: test failed
```

I googled for it for a while, but couldn't find a relevant solution..  Do you have any idea?

jeehoonkang:
@stjepang I revised this branch as you said.  Thanks!


> I feel conflicted about all this. Now we have basically downgraded to Rust from 2016. :)

Yeah..  same here.  I'd like to (1) provide a version that makes Rayon developers happy, because Rayon will be the biggest client of Crossbeam, and (2) use more recent Rusts in the future development of Crossbeam.  Maybe this PR is about (1).


> Another thing to consider is that Rayon depends on crossbeam-deque, and the deque should really switch to a lightweight hazard pointer-based scheme instead of using crossbeam-epoch. We could even implement simple HPs inside the crossbeam-deque crate. I wonder if downgrading crossbeam-epoch is actually necessary in the end. :)

Is there any evidence that a deque is better served by a hazard pointer-based scheme?  That kind of knowledge will be useful in organizing multiple garbage collection schemes in Crossbeam: separate crates for each scheme?  A single-scheme-that-rules-them-all?

ghost (approved):
Let's merge this to accommodate Rayon, but I'd like to revisit this problem later. It'd be a shame to get stuck with an old Rust version and miss out on a lot of new, useful features.

jeehoonkang:
@stjepang Thanks for the review!

Before merging it, I'd like to document and discuss the future plan.  

- As discussed in IRC, it's so much painful to be stuck in Rust 1.13.  Let's dedicate `crossbeam-epoch` 0.3 for supporting Rust 1.13, and enjoy the recent versions of Rust from `crossbeam-epoch` 0.4.  Similarly for `crossbeam-deque`.

- I'd like to merge this, publish `crossbeam-epoch` 0.3, revert this, and then publish `crossbeam-epoch` 0.4.  This is super-ugly, but it will.. work.  Similarly for `crossbeam-deque` 0.2 and 0.3.

- I think `crossbeam-epoch` 0.4 should also specify the minimum Rust version.  This is anyway a good practice :)  Let's start with the most recent stable version, and let's not fear with bumping it.

I'd like to ask what does everyone think of this tentative plan.  After reaching a consensus on a plan, I'll execute it.

jeehoonkang:
I also made a last-minute commit that removes the dependence on `NoDrop`.  For example, I changed `global: UnsafeCell<NoDrop<Arc<Global>>>` into `global: UnsafeCell<Option<Arc<Global>>>`, and `global.get().take()` just before deallocating it.  This will optimize the memory and runtime overhead a little bit (but not on par with `ManuallyDrop`).

ghost:
Isn't `NoDrop` faster than `Option`? Here's an example: when dereferencing a `NoDrop` (see its `deref` implementation) there are no checks, while `.as_ref().unwrap()` incurs the cost of a runtime check.

Also, does `Option` really have smaller memory footprint than `NoDrop`? Both are very similar enums with two variants, so I'd guess their memory overhead is the same.

jeehoonkang:
@stjepang yeah, `unwrap()` will consume few cycles..  I'm reverting the last commit.

`Option` actually consumes smaller memory, because the non-nullable optimization kicks in (I checked the LLVM IR); on the other hand, `NoDrop` prevents the optimization to happen (https://github.com/bluss/arrayvec/blob/master/nodrop/src/lib.rs#L34)

nikomatsakis:
My 2 cents on the versioning question (both for Rayon and crossbeam) is that we need to settle on rough standard across the ecosystem. I'd prefer this to be tied to an LTS-style release from Rust (i.e., we support the most recent LTS release), which would give a fixed schedule for getting access to newer stuff.

jeehoonkang:
@nikomatsakis yeah, I'm also waiting for Rust 2018 :)  The LTS release solves the problem of choosing the minimum supported Rust version for Crossbeam.

But I'm worrying about the situation that later versions of Rust will introduce an essential feature for Crossbeam.  For example, I "guess" generic associated types will be extremely helpful for dealing with the coexistence of the global and local instances of Crossbeam garbage collectors, but this feature will be [likely to be omitted from Rust 2018](https://github.com/aturon/rfcs/blob/roadmap-2018/text/0000-roadmap-2018.md#language).  Then maybe we still want to maintain two versions of Crossbeam: one for the LTS release, and another for the most recent stable version.

jeehoonkang:
I'm merging it, and executing the plan I described in https://github.com/crossbeam-rs/crossbeam-epoch/pull/61#issuecomment-360838476.  Thanks everyone for help!

PR #67, opened 2018-03-06
Allow extracting a &Collector from a Guard

Amanieu:
This is needed for https://github.com/crossbeam-rs/crossbeam-skiplist/issues/4

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -23,12 +23,11 @@ use core::ptr;
 use core::sync::atomic;
 use core::sync::atomic::Ordering;
 use alloc::boxed::Box;
-use alloc::arc::Arc;
 
 use crossbeam_utils::cache_padded::CachePadded;
 
 use atomic::Owned;
-use collector::Handle;
+use collector::{Handle, Collector};
 use epoch::{AtomicEpoch, Epoch};
 use guard::{unprotected, Guard};
 use garbage::{Bag, Garbage};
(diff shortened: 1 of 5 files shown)


src/internal.rs
```diff
@@ -188,28 +187,34 @@ impl Local {
     const PINNINGS_BETWEEN_COLLECT: usize = 128;
 
     /// Registers a new `Local` in the provided `Global`.
-    pub fn register(global: &Arc<Global>) -> Handle {
+    pub fn register(collector: &Collector) -> Handle {
```
ghost, line 190: This is a really minor nit, but would it make more sense to accept a collector by value rather than reference here?

ghost (approved):
Looks good!

Do you think it'd make sense to also expose the following methods (maybe later, not necessarily now)?

* `Handle::collector(&self) -> &Collector`
* `Guard::handle(&self) -> Option<&Handle>`

src/internal.rs
jeehoonkang: I changed it because of line 203: we need to have `collector` after making `local`.

jeehoonkang (approved):
Last year, we thought that maybe the internal API will be useful in building custom collectors.  Basically that's the reason why there are `internal.rs` and `collector.rs`.  But we couldn't find a good way to do that.

To my eyes, this PR basically says there's no longer the internal API: they exists only as the implementation details of handles and collectors.  I think it's a right direction for the current time being.

Amanieu:
I added `default_collector` to get a reference to the global collector, and `Handle::collector`.

PR #70, opened 2018-03-06
Remove impl Send for Handle

### LINKED ISSUE #65: Should `internal::Local` be `Sync`?
Currently the `internal::Local` struct is marked as `Sync`. However it uses `Cell` for the guard and handle counts, which is not `Sync`, only `Send`. I don't think it safe to mark `internal::Local` as `Sync`, but only as `Send` (given that the other field are ok to be `Send`).

ghost:
Implementation of `Send` for `Handle` was a bug. Handles really aren't supposed to be sendable to other threads.

Fixes #65.

cc @Thomasdezeeuw

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Rename `Collector::handle` to `Collector::register`.
 
+### Removed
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
This looks good to me. Maybe the changelog could use a single line mentioning that `Collector` can be shared and a `Handle` should be created per thread?

Vtec234 (approved):

Amanieu (approved):

PR #74, opened 2018-03-21
Expose the Pointer trait as public

jeehoonkang:
This PR makes the following changes:

- Mark `Pointer` as public.  This is necessary for the crates that implement data structures which expose `Atomic`-like API (e.g. `store`() and `compare_and_swap()`).

  I'm thinking of concurrent ART as a use case.  It's basically the infinite array of `Atomic<T>`, where initially everything is just `Atomic::null()`.  Basically, I'd like to expose `Art::store<P: Pointer<T>>(index: usize, value: P)`.

- Add `#![deny(missing_docs, warnings, missing_debug_implementations)]`, which is generally a good thing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,6 +58,8 @@
 #![cfg_attr(feature = "nightly", feature(alloc))]
 #![cfg_attr(not(test), no_std)]
 
+#![deny(missing_docs, warnings, missing_debug_implementations)]
+
 #[cfg(test)]
 extern crate core;
 #[cfg(all(not(test), feature = "use_std"))]
@@ -96,7 +98,7 @@ mod guard;
 mod internal;
 mod sync;
 
-pub use self::atomic::{Atomic, CompareAndSetError, CompareAndSetOrdering, Owned, Shared};
+pub use self::atomic::{Atomic, CompareAndSetError, CompareAndSetOrdering, Owned, Shared, Pointer};
 pub use self::guard::{unprotected, Guard};
 #[cfg(feature = "use_std")]
 pub use self::default::{default_collector, default_handle, is_pinned, pin};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,7 +58,7 @@
 #![cfg_attr(feature = "nightly", feature(alloc))]
 #![cfg_attr(not(test), no_std)]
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -58,6 +58,8 @@
 #![cfg_attr(feature = "nightly", feature(alloc))]
 #![cfg_attr(not(test), no_std)]
 
+#![deny(missing_docs, warnings, missing_debug_implementations)]
```
ghost: I don't feel comfortable with `deny`ing lints because future compiler updates could suddenly make our code stop compiling. See https://github.com/rust-lang/rust/pull/47204#issuecomment-356608522 for a recent example of a new warning breaking existing crates. This can be especially frustrating for crates depending on crossbeam.

We should be fine with just `#![warn(missing_docs, missing_debug_implementations)]`.

And if that is not enough, we can add `RUSTFLAGS="-D warnings"` to the Travis config. Better to break our CI builds than the whole ecosystem. :)

jeehoonkang:
@stjepang I couldn't think of that :)  I changed it to `#![warn(missing_docs, missing_debug_implementations)]`.

ghost (approved):
Since we're publicly exposing `Pointer`, this might be a good time to reconsider the naming choices around `Pointer::from_data` and `Pointer::into_data`.

The word `data` is somewhat vague and could be mistaken for the actual data behind the pointer (of type `T`). Perhaps `from_usize` and `into_usize` would sound better? Any other ideas?

jeehoonkang:
@stjepang I'm happy with `from_usize` and `into_usize`.  I initially thought `from_raw` and `into_raw`, but they probably have a connotation with raw pointers.

ghost:
@Vtec234, any thoughts on this rename?

src/lib.rs
glaebhoerl: > This can be especially frustrating for crates depending on crossbeam.

AFAIA cargo caps lints when building dependencies, so it would only affect crossbeam itself.

Vtec234 (approved):
Looks good, and I agree with the rename.

jeehoonkang:
Based on two approvals, I'm merging it.  Thanks!

PR #77, opened 2018-04-01
Remove the Garbage struct (closes #12)

### LINKED ISSUE #12: Can we think of a better name for `Garbage`?
@stjepang said:

> "Garbage" is typically in plural and refers to a bunch of deleted/dead/unlinked objects. I'd prefer to use a better synonym in singular form. That said, I don't have any ideas for a better word to use. :)

@Vtec234 replied:

> It seems that such a word literally doesn't exist: [link](https://english.stackexchange.com/questions/240876/single-countable-word-for-trash-garbage). The closest name is PieceOfGarbage.

jeehoonkang:
There was a discussion on the naming of `Garbage` (#12).  This PR simply removes it..


src/internal.rs
```diff
@@ -25,15 +44,58 @@ use core::sync::atomic::Ordering;
...
+const MAX_OBJECTS: usize = 4;
+
+/// Bag of garbages.
+#[derive(Default, Debug)]
+pub struct Bag {
+    /// Stashed objects.
+    deferreds: ArrayVec<[Deferred; MAX_OBJECTS]>,
+}
+
+unsafe impl Sync for Bag {}
```
Thomasdezeeuw: Since `Deferred` isn't `Sync` or `Send` is this ok? If so a comment explaining nice would be nice.

src/internal.rs
```diff
@@ -15,6 +15,25 @@
 //!
 //! When a participant is pinned, a `Guard` is returned as a witness that the participant is pinned.
 //! Guards are necessary for performing atomic operations, and for freeing/dropping locations.
+//!
+//! # Thread-local bag
+//!
+//! Objects that get unlinked from concurrent data structures must be stashed away until the global
```
Thomasdezeeuw, line 21: The following two paragraphs uses double spaces after a period, however the other text doesn't.

src/internal.rs
```diff
@@ -25,15 +44,58 @@ use core::sync::atomic::Ordering;
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns `true` if the bag is empty.
+    pub fn is_empty(&self) -> bool {
+        self.deferreds.is_empty()
+    }
+
+    /// Attempts to insert a garbage object into the bag and returns `true` if succeeded.
```
Thomasdezeeuw: Doc mentions `true`, but it actually returns `Ok(())`.

jeehoonkang:
@Thomasdezeeuw Thanks!  I revised the PR as you said.

Actually, `Bag` didn't need to be `Sync`.

src/internal.rs
```diff
@@ -25,15 +44,66 @@ use core::sync::atomic::Ordering;
...
+    /// Another thread may execute the `deferred` function.
+    pub unsafe fn try_push(&mut self, deferred: Deferred) -> Result<(), Deferred> {
+        self.deferreds.try_push(deferred).map_err(|e| e.element())
+    }
+}
+
+impl Drop for Bag {
+    fn drop(&mut self) {
+        // Call all deferred functions.
+        for deferred in self.deferreds.iter_mut() {
```
Vtec234: I can't see anywhere where that could happen, but are we absolutely sure a `Deferred` is never dropped as a standalone object rather than when it's inside a `Bag`? This change would then make it never run, as opposed to having the individual `Garbage` destructor take care of it before the change.

Vtec234 (approved):
Looks good, left one question.

src/internal.rs
```diff
@@ -25,15 +44,66 @@ use core::sync::atomic::Ordering;
...
+/// Bag of garbages.
+#[derive(Default, Debug)]
+pub struct Bag {
+    /// Stashed objects.
+    deferreds: ArrayVec<[Deferred; MAX_OBJECTS]>,
+}
+
+// It is safe to implement `Send` for `Bag`, since the user of `Bag::try_push()` guarantees that
+// another thread may execute any `Deferred` functions inside a `Bag`.
+unsafe impl Send for Bag {}
```
Amanieu: Wouldn't it make more sense to `impl Send` for `Deferred` instead? It will then automatically be applied to `Bag`.

src/internal.rs
jeehoonkang: I revised the doc.  Would you please review again to check if it's clearer?

src/internal.rs
jeehoonkang: I'd like to cc @stjepang.  He said that [`Deferred` is better not to implement `Send`](https://github.com/crossbeam-rs/crossbeam-epoch/issues/12#issuecomment-336735548).

src/internal.rs
Vtec234: Thank you for revising, but I think that I had something slightly different in mind. Basically, after this change, `Deferred` relies on being inside a `Bag` to be destroyed and have its stored function run. On its own, a `Deferred` won't run when dropped. I think that's fine, but just wanted to make sure.

src/internal.rs
```diff
@@ -25,15 +44,67 @@ use core::sync::atomic::Ordering;
...
 use guard::{unprotected, Guard};
-use garbage::{Bag, Garbage};
+use deferred::Deferred;
 use sync::list::{List, Entry, IterError, IsElement};
 use sync::queue::Queue;
 
+/// Maximum number of objects a bag can contain.
+#[cfg(not(feature = "strict_gc"))]
+const MAX_OBJECTS: usize = 64;
+#[cfg(feature = "strict_gc")]
```
ghost: This should be `"sanitize"`. :)

src/internal.rs
ghost: Actually, I agree with @Amanieu. Let's have `impl Send for Deferred {}`

src/internal.rs
```diff
@@ -25,15 +44,67 @@ use core::sync::atomic::Ordering;
...
+const MAX_OBJECTS: usize = 4;
+
+/// Bag of garbages.
```
ghost: Maybe change to "A bag of deferred functions"?

src/internal.rs
```diff
@@ -25,15 +44,67 @@ use core::sync::atomic::Ordering;
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns `true` if the bag is empty.
+    pub fn is_empty(&self) -> bool {
+        self.deferreds.is_empty()
+    }
+
+    /// Attempts to insert a garbage object into the bag.
```
ghost: s/garbage object/deferred function/

src/internal.rs
```diff
@@ -25,15 +44,67 @@ use core::sync::atomic::Ordering;
...
+    ///
+    /// Returns `Ok(())` if successful, and `Err(deferred)` for the given `deferred` if
+    /// unsuccessful.
```
ghost: s/unsuccessful/the bag is full/

src/internal.rs
```diff
@@ -25,15 +44,67 @@ use core::sync::atomic::Ordering;
...
+    ///
+    /// # Safety
+    ///
+    /// `deferred` should not have already been called. Also, another thread may execute the
```
ghost: I don't think this actually affects safety. If you call a deferred function twice, the second call will simply panic.

src/internal.rs
jeehoonkang: I agree with @Vtec234, so I just changed `Deferred::call()` so that it consumes `self`.  Basically `Deferred`'s API is almost the same with `FnOnce<()>`, and so we don't need to worry about double calling.

cc @stjepang, as he is the original author of `Deferred`.  May I ask why did you allow to call `Deferred` twice (and panic)?

ghost: Maybe at the time we had to be able to do `.call()` without consuming the `Deferred`? I don't remember the details.

If you can change `call` to consume `self`, go ahead. That'd be nicer than what we currently have. 👍

ghost (approved):

src/deferred.rs
```diff
@@ -19,55 +20,60 @@ pub struct Deferred {
...
+}
+
 impl Deferred {
     /// Constructs a new `Deferred` from a `FnOnce()`.
-    pub fn new<F: FnOnce()>(f: F) -> Self {
+    ///
+    /// # Safety
+    ///
+    /// It should be safe for another thread to execute the given function.
+    pub unsafe fn new<F: FnOnce()>(f: F) -> Self {
```
Thomasdezeeuw: Is there a `Send`/`Sync` for functions? Does `F: FnOnce() + Send + Sync` work, or is this already the case?

Amanieu: In this case we only need `F: FnOnce() + Send`.

Thomasdezeeuw: @Amanieu I think you're right `Sync` isn't needed.

jeehoonkang:
I made a significant change after this PR is approved.  Now `Deferred` is not `Send`, and the user of `Guard::defer()` should guarantee that it is safe for another thread to call the deferred function after the grace period.  I did it for giving a safe abstraction for unsafe implementations.

Let me explain a little bit more.  First of all, consider the following code snippet, which is the typical way of deferring the deallocation of the `shared` reference nowadays:

```rust
guard.defer(move || shared.into_owned());
```

Strictly speaking, we cannot require `F: Send` in `Deferred::new()`, because the function given above is actually not `Send`.  We also cannot require the user of `Deferred::new()` to guarantee that it is safe for another thread to call the function, because it's not the case for the function given above.  The (un)safety contract should be made in `Guard::defer()`, where the user should guarantee that it is safe for any thread to call the deferred function after the grace period.

So I made `Deferred` just an optimized container of `FnOnce()`, and let `Guard::defer()` to deal with thread safety.

I believe this change matches with @stjepang's original intent.  What do you think?

cc @Thomasdezeeuw

src/deferred.rs
```diff
@@ -52,22 +62,17 @@ impl Deferred {
                 Deferred {
                     call: call::<F>,
                     data,
+                    _marker: PhantomData,
                 }
             }
         }
     }
 
     /// Calls the function or panics if it was already called.
```
Thomasdezeeuw: Doc still mentions a panic if it was already called, but that is no longer possible.

Thomasdezeeuw (approved):

ghost:
@jeehoonkang That looks good to me, I like the latest changes!

Bikeshedding time 🚲 - if you need ideas for a fun naming scheme, here's one:

```rust
impl Bag {
    fn seal(self, epoch: Epoch) -> SealedBag {
        SealedBag { epoch, bag: self }
    }
}

struct SealedBag {
    epoch: Epoch,
    bag: Bag,
}

impl SealedBag {
    fn is_expired(self, epoch: Epoch) -> bool {
        epoch.wrapping_sub(self.epoch) >= 2
    }
}

// ...

self.queue.push(bag.seal(epoch), guard);

// ...

for _ in 0..steps {
    match self.queue.try_pop_if(
        &|sealed_bag| sealed_bag.is_expired(global_epoch),
        guard,
    ) {
        // ...
    }
}
```

But the current struct/function names are ok, too.

jeehoonkang:
@stjepang Thanks for the names.  I updated this PR as you said.

I think finally we produced a mergeable PR :)  Thanks all for your valuable comments!  I'll merge this branch once the CI becomes green.

PR #83, opened 2018-09-07
Avoid panics when exiting thread

ghost:
If we try to call `pin()` when exiting the thread, it is possible that we'll access the thread-local `HANDLE` after it has already been destroyed.

This is sometimes a problem when using `crossbeam-epoch` while the thread is unwinding from a panic. In that case, double panics cause aborts.

To fix the problem, we register a temporary `Handle` in case `HANDLE` has already been destroyed.

### DIFF
--- a/src/default.rs
+++ b/src/default.rs
@@ -20,27 +20,58 @@ thread_local! {
 /// Pins the current thread.
 #[inline]
 pub fn pin() -> Guard {
-    // FIXME(jeehoonkang): thread-local storage may be destructed at the time `pin()` is called. For
-    // that case, we should use `HANDLE.try_with()` instead.
-    HANDLE.with(|handle| handle.pin())
+    with_handle(|handle| handle.pin())
 }
 
 /// Returns `true` if the current thread is pinned.
 #[inline]
 pub fn is_pinned() -> bool {
-    // FIXME(jeehoonkang): thread-local storage may be destructed at the time `pin()` is called. For
-    // that case, we should use `HANDLE.try_with()` instead.
-    HANDLE.with(|handle| handle.is_pinned())
+    with_handle(|handle| handle.is_pinned())
 }
 
 /// Returns the default handle associated with the current thread.
 #[inline]
 pub fn default_handle() -> Handle {
-    HANDLE.with(|handle| handle.clone())
+    with_handle(|handle| handle.clone())
 }
 
 /// Returns the default handle associated with the current thread.
 #[inline]
 pub fn default_collector() -> &'static Collector {
     &COLLECTOR
+}
+
+#[inline]
+fn with_handle<F, R>(mut f: F) -> R
+where
+    F: FnMut(&Handle) -> R,
+{
+    HANDLE.try_with(|h| f(h)).unwrap_or_else(|_| f(&COLLECTOR.register()))
+}
+
+#[cfg(test)]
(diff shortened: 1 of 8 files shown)


ghost:
Note: https://github.com/servo/servo/pull/21325 is waiting on this PR.

src/default.rs
```diff
@@ -20,27 +20,58 @@ thread_local! {
...
 /// Returns the default handle associated with the current thread.
 #[inline]
 pub fn default_collector() -> &'static Collector {
     &COLLECTOR
 }
+
+#[inline]
+fn with_handle<F, R>(mut f: F) -> R
+where
+    F: FnMut(&Handle) -> R,
```
Thomasdezeeuw: Maybe FnOnce would be better, since it's only called once anyway? This might require a slight change in the code however.

src/default.rs
```diff
@@ -20,27 +20,58 @@ thread_local! {
...
+{
+    HANDLE.try_with(|h| f(h)).unwrap_or_else(|_| f(&COLLECTOR.register()))
+}
+
+#[cfg(test)]
+mod tests {
+    use crossbeam_utils::thread;
+
+    #[test]
+    fn pin_while_exiting() {
```
Thomasdezeeuw, line 50: Is this only meant to not abort? If so a comment explaining that would be useful, now the test doesn't really make much sense.

Thomasdezeeuw (approved):

src/default.rs
ghost: Unfortunately, that doesn't compile and I don't know how to do better.

```
error[E0382]: capture of moved value: `f`
  --> src/default.rs:49:50
   |
49 |     HANDLE.try_with(|h| f(h)).unwrap_or_else(|_| f(&COLLECTOR.register()))
   |                     ---                          ^ value captured here after move
   |                     |
   |                     value moved (into closure) here
   |
   = note: move occurs because `f` has type `F`, which does not implement the `Copy` trait
```

Thomasdezeeuw: Sorry I thought that it might work with a match statement, however after trying it I see that it doesn't work.

Thomasdezeeuw: But I think you can pass `f` directly to `try_with` right? It doesn't need an extra closure.

ghost: We need an extra closure. :) Otherwise we get this:

```
error[E0382]: capture of moved value:
  --> src/default.rs:49:39
   |
49 |     HANDLE.try_with(f).unwrap_or_else(|_| f(&COLLECTOR.register()))
   |                     -                 ^^^ value captured here after move
   |                     |
   |                     value moved here
   |
   = note: move occurs because `f` has type `F`, which does not implement the `Copy` trait
```

Vtec234 (approved):

src/default.rs
```diff
@@ -20,27 +20,61 @@ thread_local! {
...
-    // that case, we should use `HANDLE.try_with()` instead.
-    HANDLE.with(|handle| handle.is_pinned())
+    with_handle(|handle| handle.is_pinned())
 }
 
 /// Returns the default handle associated with the current thread.
 #[inline]
 pub fn default_handle() -> Handle {
-    HANDLE.with(|handle| handle.clone())
+    with_handle(|handle| handle.clone())
```
Vtec234: This is removed in #84.

PR #85, opened 2018-09-08
Add defer_unchecked and defer_destroy

### LINKED ISSUE #75: Guard::defer is a footgun
@Amanieu just hit an annoying bug where he forgot to write `move` before the closure:

```rust
guard.defer(|| Self::finalize(self));
```

The documentation of `defer` suggests **always** using `move`, but it's too easy to forget about this rule and run into very difficult to find bugs. Clearly, we should do something here.

My suggestion is to have two variants of `defer`, a checked and an unchecked version (similar to `get` vs `get_unchecked`, or `unreachable!()` vs `unreachable_unchecked()`):

```rust
fn defer<F, R>(&self, f: F)
where
    F: FnOnce() -> R + Send + 'static;

unsafe fn defer_unchecked<F, R>(&self, f: F)
where
    F: FnOnce() -> R;
```

That means the original call to `defer` would be written as:

```rust
let ptr = self as *const _ as usize;
guard.defer(|| Self::finalize(ptr as *const Self));
```

Or, using the more dangerous option:

```rust
guard.defer_unchecked(move || Self::finalize(self));
```

This situation reminds me of [`spawn_unsafe`](https://docs.rs/crossbeam/0.3.2/crossbeam/fn.spawn_unsafe.html) (it should really be named `spawn_unchecked`), which is equivalent to `std::thread::spawn`, except it doesn't require `F: 'static` (and the `F: Send` bound should probably be lifted too, IMO; see [this](https://github.com/crossbeam-rs/crossbeam/issues/54)).

ghost:
Now `defer` requires `F: Send + 'static`, while `defer_unchecked` doesn't have the 

We also add `defer_destroy` which is more convenient and less dangerous than `defer_unchecked(move || s.into_owned())`.

Closes #75.

### DIFF
--- a/src/guard.rs
+++ b/src/guard.rs
@@ -88,9 +88,10 @@ impl Guard {
     /// functions from both local and global caches may get executed in order to incrementally
     /// clean up the caches as they fill up.
     ///
-    /// There is no guarantee when exactly `f` will be executed. The only guarantee is that won't
-    /// until all currently pinned threads get unpinned. In theory, `f` might never be deallocated,
-    /// but the epoch-based garbage collection will make an effort to execute it reasonably soon.
+    /// There is no guarantee when exactly `f` will be executed. The only guarantee is that it
+    /// won't be executed until all currently pinned threads get unpinned. In theory, `f` might
+    /// never run, but the epoch-based garbage collection will make an effort to execute it
+    /// reasonably soon.
     ///
     /// If this method is called from an [`unprotected`] guard, the function will simply be
     /// executed immediately.
@@ -114,9 +115,10 @@ impl Guard {
     /// functions from both local and global caches may get executed in order to incrementally
     /// clean up the caches as they fill up.
     ///
-    /// There is no guarantee when exactly `f` will be executed. The only guarantee is that won't
-    /// until all currently pinned threads get unpinned. In theory, `f` might never be deallocated,
-    /// but the epoch-based garbage collection will make an effort to execute it reasonably soon.
+    /// There is no guarantee when exactly `f` will be executed. The only guarantee is that it
+    /// won't be executed until all currently pinned threads get unpinned. In theory, `f` might
+    /// never run, but the epoch-based garbage collection will make an effort to execute it
+    /// reasonably soon.
     ///
     /// If this method is called from an [`unprotected`] guard, the function will simply be
     /// executed immediately.
@@ -213,9 +215,9 @@ impl Guard {
     /// incrementally clean up the caches as they fill up.
     ///
     /// There is no guarantee when exactly the destructor will be executed. The only guarantee is
-    /// that won't until all currently pinned threads get unpinned. In theory, `f` might never be
-    /// deallocated, but the epoch-based garbage collection will make an effort to execute it
-    /// reasonably soon.
+    /// that it won't be executed until all currently pinned threads get unpinned. In theory, the
+    /// destructor might never run, but the epoch-based garbage collection will make an effort to
+    /// execute it reasonably soon.
     ///
     /// If this method is called from an [`unprotected`] guard, the destructor will simply be
     /// executed immediately.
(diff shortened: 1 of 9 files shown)


jeehoonkang (approved):

src/guard.rs
```diff
@@ -79,6 +80,32 @@ pub struct Guard {
...
 impl Guard {
+    /// Stores a function so that it can be executed at some point after all currently pinned
+    /// threads get unpinned.
+    ///
+    /// This method first stores `f` into the thread-local (or handle-local) cache. If this cache
+    /// becomes full, some functions are moved into the global cache. At the same time, some
+    /// functions from both local and global caches may get executed in order to incrementally
+    /// clean up the caches as they fill up.
+    ///
+    /// There is no guarantee when exactly `f` will be executed. The only guarantee is that won't
```
Vtec234: `won't` -> `it won't execute`

src/guard.rs
```diff
@@ -79,6 +80,32 @@ pub struct Guard {
...
+    /// until all currently pinned threads get unpinned. In theory, `f` might never be deallocated,
```
Vtec234: `never be deallocated` -> `never run`

src/guard.rs
```diff
@@ -106,8 +133,8 @@ impl Guard {
     /// let guard = &epoch::pin();
     /// let message = "Hello!";
     /// unsafe {
-    ///     // ALWAYS use `move` when sending a closure into `defef`.
```
Vtec234, line 109: Despite the constant negative press defefe? :P

src/guard.rs
```diff
@@ -177,6 +204,80 @@ impl Guard {
...
+    /// Stores a destructor for an object so that it can be deallocated and dropped at some point
+    /// after all currently pinned threads get unpinned.
+    ///
+    /// This method first stores the destructor into the thread-local (or handle-local) cache. If
+    /// this cache becomes full, some destructors are moved into the global cache. At the same
+    /// time, some destructors from both local and global caches may get executed in order to
+    /// incrementally clean up the caches as they fill up.
+    ///
+    /// There is no guarantee when exactly the destructor will be executed. The only guarantee is
+    /// that won't until all currently pinned threads get unpinned. In theory, `f` might never be
```
Vtec234: `f might never` -> `ptr might never`

src/guard.rs
```diff
@@ -177,6 +204,80 @@ impl Guard {
...
+    /// deallocated, but the epoch-based garbage collection will make an effort to execute it
```
Vtec234: `execute it` -> `drop it`

src/guard.rs
```diff
@@ -177,6 +204,80 @@ impl Guard {
...
+    ///
+    /// If this method is called from an [`unprotected`] guard, the destructor will simply be
+    /// executed immediately.
+    ///
+    /// # Safety
+    ///
+    /// The object must not be reachable by other threads anymore, otherwise it might be still in
+    /// use when the destructor runs.
+    ///
+    /// Apart from that, keep in mind that another thread may execute `f`, so the object must be
```
Vtec234: `may execute f` -> `may execute the destructor`

src/guard.rs
```diff
@@ -177,6 +204,80 @@ impl Guard {
...
+    /// sendable to other threads.
+    ///
+    /// We intentionally didn't require `F: Send`, because Rust's type systems usually cannot prove
```
Vtec234: `F: Send` -> `T: Send` (also on line below)

src/guard.rs
```diff
@@ -177,6 +204,80 @@ impl Guard {
...
+    /// `F: Send` for typical use cases. For example, consider the following code snippet, which
+    /// exemplifies the typical use case of deferring the deallocation of a shared reference:
+    ///
+    /// ```ignore
+    /// let shared = Owned::new(7i32).into_shared(guard);
+    /// guard.defer_destroy(move || shared.into_owned()); // `Shared` is not `Send`!
```
Vtec234: `move || shared.into_owned()` -> `shared`

Vtec234 (approved):
Looks good, small comment fixes needed. Also, it's funny that in the end we're going back to parts of the old design for deferred drop :)

PR #87, opened 2018-10-31
[WIP] Support DSTs

jeehoonkang:
Currently, crossbeam-epoch doesn't support dynamically-sized types (DST): https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-05-02-atomic-api.md#fat-pointers-and-dsts

This PR adds the support for DST by generalizing `Box<T>`, which was used as the underlying storage type for atomic pointers, to the `Storage<T>` trait.  For more details, see the comment on the `Storage` trait.  As an example, I added support for [dynamically-sized arrays](https://github.com/crossbeam-rs/crossbeam-epoch/pull/87/files#diff-79833cfe519a852b29369ee44aa044acR10).  Other kinds of DSTs can be supported by implementing `Storage<T>`, even in the client code.

This PR was motivated by the fact that there's an unnecessary indirection in `crossbeam-deque` when accessing its underlying buffer.  As a proof of concept, I removed that indirection on top of PR in my local development: [`concurrent-circbuf`](https://github.com/jeehoonkang/concurrent-circbuf/blob/master/src/sp/mod.rs#L77).

This is a **breaking change**, but I tried to keep it as small as possible.  As far as I understand, the only breaking change is renaming `Atomic::from(*const T)` into `Atomic::from_raw(*const T)` and `Shared::from(*const T)` into `Shared::from_raw(*const T)`.  (Question: I ended up changing the underlying types.  Is it a breaking change?)

This is a WIP.  Probably we want to discuss this PR's design and even write up an RFC.  I just share it for starting a discussion.  Thanks!


src/storage/box_storage.rs
```diff
@@ -0,0 +1,100 @@
...
+use atomic::decompose_data;
+
+use {AtomicTmpl, OwnedTmpl, SharedTmpl, Pointer, Storage};
+
+/// An atomic pointer that can be safely shared between threads.
+///
+/// See [`AtomicTmpl`] for more details.
+///
+/// [`AtomicTmpl`]: struct.AtomicTmpl.html
+pub type Atomic<T> = AtomicTmpl<T, Box<T>>;
```
jeehoonkang, line 12: I wonder if it's a breaking change.  `Atomic` is renamed into `AtomicTmpl`, with an additional type parameter, and `Atomic` is now defined as a type alias.

jeehoonkang:
Closing as duplicate: https://github.com/crossbeam-rs/crossbeam/pull/209

